PR #11, opened 2020-01-07
AS-139: Add import status endpoints

MatthewBemis:


helgridly (changes requested):
buncha nits and style things but this is good!

davidangb (commented):
some comments inline

davidangb (commented):
two more comments ...

davidangb (approved):
:+1: lgtm, and I defer to @helgridly for python best practices

helgridly (approved):
i have a couple final comments but i'll thumb pending them

PR #13, opened 2020-01-13
[AS-136] Translate from PFB to Rawls batchUpsert

helgridly:
I tested with 14040 cases (`Carotid Plaque: plaque not present`) and it successfully created the batchUpsert file (even with the default F1 memory limit of 256MB!).


app/auth/service_auth.py
```diff
@@ -20,56 +20,65 @@
 ]
 
 
-class TokenAndExpiry(NamedTuple):
-    token: str
-    expiry: datetime.datetime
+class CachedCreds:
```
helgridly, line 24: this is a reshuffle of existing code to make both the token and the google `Credentials` object available. it does pretty much the same thing

app/server/requestutils.py
```diff
@@ -0,0 +1,103 @@
+import logging
```
helgridly, line 1: this used to be `httputils.py`, I moved it so we could have `server.requestutils` for incoming requests, and `util.http` for outgoing ones

app/server/requestutils.py
```diff
@@ -0,0 +1,103 @@
...
+            # Return a 500 and a UUID which developers can look up in the log.
+            # NOTE: This will log callstack information and potentially user values.
+            eid = uuid.uuid4()
+            logging.error(f"eid {eid}:\n{traceback.format_exc()}")
+            return flask.make_response(f"Internal Server Error\nerror id: {eid}", 500)
+
+    return catch_excs
+
+
+def pubsubify_excs(some_func: Callable[..., flask.Response]):
```
helgridly, line 41: this is new (and important)

app/service.py
```diff
@@ -67,4 +68,4 @@ def handle(request: flask.Request) -> flask.Response:
 
     pubsub.publish({"action": "translate", "import_id": new_import_id})
 
-    return flask.make_response((str(new_import_id), 200))
+    return flask.make_response((str(new_import_id), 201))
```
helgridly, line 70: this should have been 201 as it's a new thing

app/status.py
```diff
@@ -42,7 +42,7 @@ def handle_list_import_status(request: flask.Request, ws_ns: str, ws_name: str)
         q = sess.query(model.Import).\
             filter(model.Import.workspace_namespace == ws_ns).\
             filter(model.Import.workspace_name == ws_name)
-        q = q.filter(model.Import.status == ImportStatus.Running) if running_only else q
+        q = q.filter(model.Import.status.in_(ImportStatus.running_statuses())) if running_only else q
```
helgridly, line 43: quick fix to a previous PR: there is more than one running status

app/tests/test_path_validation.py
```diff
@@ -31,60 +31,60 @@ def assert_response_code_and_logs(resp, caplog, import_url):
...
+    resp = client.post('/iservice/namespace/name/imports', json=payload, headers=good_headers)
+    assert resp.status_code == 201
+    # NB we don't test anything deeper than the 201 response; other tests check to see if the
     # db is updated, etc.
 
 @pytest.mark.usefixtures(sam_valid_user, user_has_ws_access, pubsub_publish, "pubsub_fake_env")
 def test_illegal_netloc_simple(client: flask.testing.FlaskClient, caplog):
     payload = {"path": f"https://haxxor.evil.bad/some/valid/path", "filetype": "pfb"}
-    resp = client.post('/iservice/namespace/name/import', json=payload, headers=good_headers)
+    resp = client.post('/iservice/namespace/name/imports', json=payload, headers=good_headers)
```
helgridly, line 42: renames from the merge

app/tests/test_requestutils.py
```diff
@@ -0,0 +1,54 @@
+import pytest
```
helgridly, line 1: this used to be test_httputils

app/tests/test_requestutils.py
```diff
@@ -0,0 +1,54 @@
...
+
+
+def test_pubsubify_excs(client: flask.testing.FlaskClient):
+    # pre-populate an import that will get error'd
+    with session_ctx() as sess:
+        new_import = model.Import("aa", "aa", "uuid", "aa@aa.aa", "gs://aa/aa", "pfb")
+        sess.add(new_import)
+        sess.commit()
+
+    @pubsubify_excs
```
helgridly, line 43: this test is new

app/tests/test_status.py
```diff
@@ -24,11 +24,13 @@
...
 def test_get_import_status(client):
-    import_id = client.post('/iservice/namespace/name/imports', json=good_json, headers=good_headers).get_data(as_text=True)
+    new_import_resp = client.post('/iservice/namespace/name/imports', json=good_json, headers=good_headers)
+    assert new_import_resp.status_code == 201
+    import_id = new_import_resp.get_data(as_text=True)
 
     resp = client.get('/iservice/namespace/name/imports/{}'.format(import_id), headers=good_headers)
     assert resp.status_code == 200
-    assert resp.get_data(as_text=True) == json.dumps({'id': import_id, 'status': ImportStatus.Pending.name})
+    assert resp.get_json(force=True) == {'id': import_id, 'status': ImportStatus.Pending.name}
```
helgridly, line 33: converting the response into a python object is better than comparing to the stringified version of that object, in case the stringification is different with whitespace/tabs or something

app/tests/test_status.py
```diff
@@ -45,29 +47,29 @@ def test_get_all_import_status(client):
...
     resp = client.get('/iservice/namespace/name/imports', headers=good_headers)
     assert resp.status_code == 200
-    assert resp.get_data(as_text=True) == json.dumps([{"id": import_id, "status": ImportStatus.Pending.name}])
+    assert resp.get_json(force=True) == [{"id": import_id, "status": ImportStatus.Pending.name}]
 
 
 @pytest.mark.usefixtures(sam_valid_user, user_has_ws_access, pubsub_publish, "pubsub_fake_env")
 def test_get_all_running_when_none(client):
-    client.post('/iservice/namespace/name/imports', json=good_json, headers=good_headers)
+    # poke in one import that's in the Done state
```
helgridly, line 55: this is a fix for the "multiple running statuses" mentioned earlier

app/translate.py
```diff
@@ -1,24 +1,67 @@
 import flask
 import logging
 
-from typing import Dict, Optional
+from app.auth import service_auth
```
helgridly, line 4: hooray, you finally got to the meat of this PR!

app/translators/pfb_to_rawls.py
```diff
@@ -0,0 +1,66 @@
+import base64
+from app.translators.translator import Translator
+from pfb.reader import PFBReader
+from typing import Iterator, Dict, Set, Tuple, Any
+
```
helgridly, line 5: this is lifted straight from [avro-import](https://github.com/broadinstitute/avro-import/blob/develop/function/main.py)

app/translators/pfb_to_rawls.py
```diff
@@ -0,0 +1,66 @@
...
+            options = {}
+        defaults = {'b64-decode-enums': False, 'prefix-object-ids': False}
+        self.options = {**defaults, **options}
+
+    def translate(self, file_like) -> Iterator[Dict[str, Any]]:
+        with PFBReader(file_like) as reader:
+            schema = reader.schema
+            enums = self.list_enums(schema)
+            result_gen = (self.translate_record(record, enums)
+                          for record in reader if record['name'] != 'Metadata')
```
helgridly: small change here: this returns a generator, not the actual list, in order to stream out of http, through the translator, and into the gs bucket

app/tests/conftest.py
```diff
@@ -13,18 +13,31 @@
 import sqlalchemy.orm
 
 from app import create_app
-from app.db import db, model
+from app.db import db, model, DBSession
 
 
-@pytest.fixture(scope="session")
-def client() -> flask.testing.FlaskClient:
+def _test_client() -> flask.testing.FlaskClient:
```
helgridly, line 22: changes here fix a nasty bug in tests. the test function `test_pubsubify_excs` adds a new URL to the client to test that decorator. but flask throws an error if you do that after you've handled your first request. but moving `test_httputils` to `test_requestutils` changed the order of test runs, which meant it went from being called first (thus not triggering the flask error) to not-first (meaning it did). the creation of the new fixture scoped at `"session"` (below) means that test functions that want to do shenanigans like this can do so and stay isolated from everyone else.

app/translate.py
```diff
@@ -1,24 +1,67 @@
...
+def translate(msg: Dict[str, str]) -> flask.Response:
+    import_id = msg["import_id"]
+    with db.session_ctx() as sess:
+        # flip the status to Translating, and then get the row
+        update_successful = Import.update_status_exclusively(import_id, ImportStatus.Pending, ImportStatus.Translating, sess)
+        import_details: Import = Import.reacquire(import_id, sess)
+
+    if not update_successful:
+        # this import wasn't in pending. most likely this means that the pubsub message we got was delivered twice,
+        # and some other GAE instance has picked it up and is happily processing it. happy translating, friendo!
```
davidangb, line 38: let's at least log this case, so we can monitor how frequently this happens in practice.  I'd love to distinguish the "id not found" and "status not in pending" cases, too - which as written  could only happen inside `update_status_exclusively`

app/translate.py
```diff
@@ -1,24 +1,67 @@
...
+FILETYPE_TRANSLATORS = {"pfb": PFBToRawls}
 
 VALID_NETLOCS = ["gen3-pfb-export.s3.amazonaws.com", "storage.googleapis.com"]
 
-def translate(attributes: Dict[str, str]) -> flask.Response:
-    # the filetype attribute has been validated on ingest, so we don't need to revalidate it here.
-    # at some point in the future this function will flesh out to ingest other filetypes.
-    return pfb_to_rawls(attributes)
+def translate(msg: Dict[str, str]) -> flask.Response:
+    import_id = msg["import_id"]
```
davidangb, line 30: should be very safe, but worth checking to see if `import_id` exists inside the message. I guess if it doesn't, `update_status_exclusively` will fail, so maybe we don't need anything else.

app/util/exceptions.py
```diff
@@ -1,13 +1,18 @@
...
 class ISvcException(Exception):
-    def __init__(self, message: str, http_status: int = 500, audit_logs: Optional[List[str]] = None):
+    def __init__(self, message: str, http_status: int = 500, imports: Optional[List[Import]] = None, audit_logs: Optional[List[str]] = None, retry_pubsub: bool = False):
         if audit_logs is None:
             audit_logs = []
         self.message = message
         self.http_status = http_status
         self.audit_logs = audit_logs
+        self.retry_pubsub = retry_pubsub
+        self.imports = imports if imports else []
```
davidangb, line 21: would you mind updating the `audit_logs` line to use this syntax, for consistency?

app/util/json.py
```diff
@@ -0,0 +1,27 @@
+
+class StreamArray(list):
+    """
+    Converts a generator into a list object that can be json serialisable
+    while still retaining the iterative nature of a generator.
+
+    IE. It converts it to a list without having to exhaust the generator
+    and keep it's contents in memory.
```
davidangb: nit: "its"

davidangb (commented):
minor comments inline, I wish I had more to say on the python-ness of this PR.

I agree this needs more tests. Potentially we could copy https://github.com/uc-cdis/pypfb/tree/master/tests/pfb-data to use as test data.

app/translators/pfb_to_rawls.py
```diff
@@ -0,0 +1,71 @@
...
+            schema = reader.schema
+            enums = self.list_enums(schema)
+
+            # You might think that the following is equivalent:
+            # (self.translate_record(record,enums) for record in reader if record['name'] != 'Metadata')
+            # But that fails, and this doesn't, possibly because something upstream is reading the generator
+            # more than once.
+            for record in reader:
+                if record['name'] != 'Metadata':
+                    yield self.translate_record(record, enums)
```
helgridly, line 25: ...with a minor change here to make this return a generator, not a list

app/tests/test_model.py
```diff
@@ -0,0 +1,38 @@
...
+
+
+def test_reacquire(fake_import: model.Import):
+    with db.session_ctx() as sess:
+        sess.add(fake_import)
+
+    with db.session_ctx() as sess2:
+        reacquired: model.Import = model.Import.reacquire(fake_import.id, sess2)
+        print(fake_import)
+        print(reacquired)
```
davidangb: stray `print`s?

app/translate.py
```diff
@@ -1,25 +1,94 @@
...
+        raise exceptions.SystemException([import_details], e)
+    except Exception as e:
+        # Something went wrong with the translate. Raising an exception will fail the import.
+        # Over time we should be able to narrow down the kinds of exception we might get, and perhaps
+        # give users clearer messaging instead of logging them all.
+        # For now, this is a last-ditch catch-all.
+        print("file xlate exc")
+        raise exceptions.FileTranslationException(import_details, e)
+
+    print("xlate success")
```
davidangb: stray `print`s, here and in the two exception cases previous

app.yaml.ctmpl
```diff
@@ -10,10 +10,14 @@ env_variables:
...
+    BATCH_UPSERT_BUCKET: "import-service-batchupsert-{env}"
 
     PUBSUB_PROJECT: "{env.project}"
     PUBSUB_TOPIC: "task_chunk_topic"
     PUBSUB_SUBSCRIPTION: "task_chunk_subscription"
     PUBSUB_TOKEN: "{secrets.pubsub_token}"
     PUBSUB_ACCOUNT: "{secrets.pubsub_account}"
-    PUBSUB_AUDIENCE: "{secrets.pubsub_audience}"
\ No newline at end of file
+    PUBSUB_AUDIENCE: "{secrets.pubsub_audience}"
```
ansingh7115, line 20: I've never used pub sub audiences before - is there anything that's required on the rawls end for this?

helgridly: AFAICT from the documentation, the audience is added by PubSub, so nothing for you to do. We should coordinate to test our branches work together and make sure e.g. that Rawls has access to push to the status topic.

app/translate.py
helgridly: If the import service is getting a `translate` message on its pubsub referring to an import that doesn't exist, something's gone really wrong, given that it's the only one who puts that message on its queue. Also it'd throw a SQLAlchemy exception in the `reacquire` call because it calls [Query.one()](https://docs.sqlalchemy.org/en/13/orm/query.html#sqlalchemy.orm.query.Query.one) under the hood

app/db/model.py
```diff
@@ -23,13 +44,33 @@ def __init__(self, enum: Type[T]) -> None: ...
 @enum.unique
 class ImportStatus(enum.Enum):
     Pending = enum.auto()
-    Running = enum.auto()
+    Translating = enum.auto()
+    Upserting = enum.auto()
+    Error = enum.auto()
+    Done = enum.auto()
```
ansingh7115: I have the status "Ready For Upsert" in rawls - does that correlate  to "Pending" here?

app/translate.py
helgridly: The code as written would throw a `KeyError` on this line, log the error, ack the pubsub message, and then any associated import would sit in `Pending` forever. We could check, but without an `import_id` we wouldn't be able to report an error to the user, so in the end the only difference would be the message that ends up in the log.

app/db/model.py
helgridly: No, we need to sync on this. I can make them `ReadyForUpsert` and `Upserting` and fix the code my side. `Pending` is "we received the import request from the user but haven't done anything at all with it yet".

deployall.sh
```diff
@@ -16,6 +16,15 @@ gcloud pubsub subscriptions create task_chunk_subscription \
...
     --ack-deadline 10
 
+
+gcloud pubsub subscriptions create task_chunk_subscription \
+    --topic task_chunk_topic \
+    --push-endpoint \
+    "https://import-service-dot-broad-dsde-dev.appspot.com/_ah/push-handlers/receive_messages?token=$(cat token.secret)" \
+    --ack-deadline 10 \
+    --push-auth-service-account="import-service-pubsub@broad-dsde-dev.iam.gserviceaccount.com" \
+    --push-auth-token-audience="importservice.dev.test.firecloud.org"
```
ansingh7115, line 20: should the service account / audience values be in config?

deployall.sh
```diff
@@ -16,6 +16,15 @@ gcloud pubsub subscriptions create task_chunk_subscription \
     "https://import-service-dot-broad-dsde-dev.appspot.com/_ah/push-handlers/receive_messages?token=$(cat token.secret)" \
     --ack-deadline 10
 
+
+gcloud pubsub subscriptions create task_chunk_subscription \
+    --topic task_chunk_topic \
```
ansingh7115: Idk if `task_chunk_topic` is a good name - something along the lines of `terra-async-import-topic` might be better

helgridly: this is just a crappy deploy script for us to use while testing, it won't be our real deploy process. that's covered by AS-172; we can make sure things are sensibly named, config'd, and consistent across all envs when we get there

app/auth/service_auth.py
```diff
@@ -16,60 +16,70 @@
 
 IMPORT_SERVICE_SCOPES = [
     "https://www.googleapis.com/auth/userinfo.email",
-    "https://www.googleapis.com/auth/userinfo.profile"
+    "https://www.googleapis.com/auth/userinfo.profile",
+    "https://www.googleapis.com/auth/devstorage.full_control"
```
davidangb: principle of least privilege: would this work with `devstorage.read-write` instead?

app/db/model.py
```diff
@@ -50,7 +94,26 @@ def __init__(self, workspace_name: str, workspace_ns: str, workspace_uuid: str,
...
+            .where(Import.status == current_status) \
+            .values(status=new_status)
+        num_affected_rows = sess.execute(update).rowcount
+        return num_affected_rows > 0
 
-    def __repr__(self):
-        # todo: replace with https://github.com/manicmaniac/sqlalchemy-repr
-        return f"<Import({self.id}, {self.workspace_name}, {self.workspace_namespace}, {self.submitter}, {self.submit_time}, {self.status})>"
+    def write_error(self, msg: str) -> None:
+        self.error_message = msg
```
davidangb, line 128: truncate to 2048 chars to ensure we don't overflow the db column? 2048 is a lot of chars for an error message, but I'm envisioning a corner case where the `msg` contains a full stack trace or something like that.

app/server/requestutils.py
```diff
@@ -0,0 +1,105 @@
...
+    """Takes a pattern like "/foo/<boo>/woo" and tests the request path against it.
+    Returns {"boo": something} only if the request path matches the pattern and has no slashes in it."""
+    regex = _pattern_to_regex(pattern)
+
+    m = re.match(regex, request_path)
+    if m is None:
+        logging.info(f"couldn't match {request_path} against {pattern}")
+        raise NotFoundException()
+    else:
+        return m.groupdict()
```
davidangb, line 105: I know this is older code that already passed review, but I can't help thinking ... isn't there a library for this somewhere? Do we have to write our own?

app/tests/test_translate.py
```diff
@@ -0,0 +1,209 @@
...
+    @memunit.assert_gt_mb(current_memusage_mb + 10)
+    def bad_noop_translate():
+        maybe_himem_work(tmp_path / "numbers.txt", BadNoOpTranslator())
+
+    # Test that a streamy translator stays under the memory limit.
+    good_noop_translate()
+
+    # A bad, in-memory translator should go OVER the memory limit. If this test fails, then the numbers-file
+    # isn't big enough to accurately test whether our streaming works.
+    bad_noop_translate()
```
davidangb, line 66: this is a super cool test, and I'm also simultaneously a bit worried about measuring memory usage, especially in a Circle (Travis, etc.) environment where we don't have full control. No action necessary - but if we find this test to be flaky, we'll have to find a different  approach.

davidangb (commented):
hopefully not-big comments inline

app/server/requestutils.py
helgridly: this is leftover from the old cloud function version of import service. flask can do this for you -- i'm going to do that work after this ticket is resolved

app/db/model.py
helgridly: great call

app/tests/test_translate.py
helgridly: agreed. i figured that since "don't run out of memory" is such an important requirement for import service it was probably worth testing for reals. in the absence of integration tests this seemed like an okayish place to do it

app/auth/service_auth.py
helgridly: probably -- i'll give it a try, thanks

ansingh7115 (approved):

davidangb (approved):
:+1: assuming you address the 2048-char `msg` limit and can use `devstorage.read-write`

PR #16, opened 2020-02-05
[noticket] A few improvements that have been bugging me for a while

helgridly:
* Use Flask parsing for url parameters instead of hand-rolling our own `requestutils.expect_urlshape`
* Remove the old `chunk_task` module, which was used for testing pubsub
* Remove the redundant `/iservice` prefix which is no longer necessary now we're off Cloud Functions
* Clean up database setup/teardown in tests, and fix a bug where raising an exception inside a database session would nuke all data in your test function


app/db/db.py
```diff
@@ -13,16 +13,12 @@
 db_name = os.environ.get("DB_NAME")
 cloud_sql_connection_name = os.environ.get("CLOUD_SQL_CONNECTION_NAME")
 
-# Store the db so it can be reused between Cloud Function invocations.
-# Storing the session seems sketchier, and results in strange behaviour if application code isn't
-# careful to close its sessions. Application code should use session_ctx() to make sure this is handled
-# cleanly.
+# Store the db so it can be reused between GAE invocations.
```
helgridly, line 16: Attempting to reuse the session is silly. They're not expensive, complicates the test harness, and could potentially have led to weirdness when reusing GAE VMs.

app/db/model.py
```diff
@@ -108,7 +108,7 @@ def __init__(self, workspace_name: str, workspace_ns: str, workspace_uuid: str,
         self.error_message = None
 
     @classmethod
-    def reacquire(cls, id: str, sess: DBSession) -> ImportT:
+    def get(cls, id: str, sess: DBSession) -> ImportT:
```
helgridly, line 111: `reacquire` was a bad name.

app/new_import.py
```diff
@@ -28,19 +27,15 @@
 schema_validator = jsonschema.Draft7Validator(NEW_IMPORT_SCHEMA)
 
 
-def handle(request: flask.Request) -> flask.Response:
-    request_path = request.path
-
-    urlparams = requestutils.expect_urlshape('/iservice/<ws_ns>/<ws_name>/imports', request_path)
-
+def handle(request: flask.Request, ws_ns: str, ws_name: str) -> flask.Response:
```
helgridly, line 30: function is cleaner now we're not doing our own URL parsing

app/server/requestutils.py
```diff
@@ -75,31 +75,3 @@ def catch_excs(*args, **kwargs) -> flask.Response:
             return flask.make_response(f"Internal Server Error\nerror id: {eid}", PUBSUB_STATUS_NOTOK)  # don't retry mystery errors
 
     return catch_excs
-
```
helgridly, line 78: unnecessary when you use Flask URL parsing

app/server/routes.py
```diff
@@ -2,28 +2,28 @@
...
-from app import service, translate, status
+from app import new_import, translate, status
 import app.auth.service_auth
 from app.server.requestutils import httpify_excs, pubsubify_excs
 
 routes = flask.Blueprint('import-service', __name__, '/')
 
 
-@routes.route('/iservice/<path:rest>', methods=["POST"])
+@routes.route('/<ws_ns>/<ws_name>/imports', methods=["POST"])
```
helgridly, line 12: * nix `/iservice` prefix
* use flask URL parsing

app/tests/conftest.py
```diff
@@ -47,33 +47,28 @@ def client_with_modifiable_routes() -> flask.testing.FlaskClient:
...
 
-    model.Base.metadata.create_all(_db)
+    # Monkeypatch over the autogenerated database with a sqlalchemy one.
+    # It will get its tables recreated and dropped around every test function.
+    db._db = _db
 
-    _connection = _db.connect()
-    yield _connection
-    _connection.close()
+    yield _db
```
helgridly, line 59: This is much less yikes than the YIKES!!!! in what used to be `def dbsession` below.

app/tests/conftest.py
```diff
@@ -47,33 +47,28 @@ def client_with_modifiable_routes() -> flask.testing.FlaskClient:
...
-    txn.rollback()
+def dbconnection(_db_internal: sqlalchemy.engine.Engine) -> Iterator[sqlalchemy.engine.Connection]:
+    """"""
+    model.Base.metadata.create_all(_db_internal)
+    _connection = _db_internal.connect()
+
+    yield _connection
+
+    _connection.close()
+    model.Base.metadata.drop_all(_db_internal)
```
helgridly, line 71: Instead of running test functions in a transaction, just create and destroy the tables each time. We use a sqlite in-memory database, so it's fast.

MatthewBemis (approved):
lgtm

ansingh7115 (approved):
Looks good

PR #17, opened 2020-02-06
[AS-137] Pubsub message handing to accept import status updates from trusted external services

helgridly:


app/tests/conftest.py
```diff
@@ -16,7 +16,7 @@
 
 from app import create_app
 from app.auth import service_auth
-from app.db import db, model, DBSession
+from app.db import db, model
```
helgridly, line 19: stuff in this file is unrelated to the pubsub message handling, i just poked it in because i noticed it. it's just cruft removal / tidying

app/status.py
```diff
@@ -45,3 +47,36 @@ def handle_list_import_status(request: flask.Request, ws_ns: str, ws_name: str)
...
+    update_successful = True
+    with db.session_ctx() as sess:
+        imp: model.Import = model.Import.get(import_id, sess)
+
+        # Quick summary:
+        #   If the import is in a terminal status, the caller did something bad.
+        #   If the caller is setting to error, ignore current status and jump straight there.
+        #   Otherwise update the status if the caller got the previous one correct.
+        if imp.status in ImportStatus.terminal_statuses():
+            raise exceptions.TerminalStatusChangeException(import_id, new_status, imp.status)
```
davidangb: I wonder if we'll hit this due to pubsub's at-least-once promise. If we get two messages marking an import as Done,  this will throw an exception. We could avoid that by checking if `new_status` and `imp.status` are equal and calling that a noop (log it at info level for visibility)

app/status.py
```diff
@@ -45,3 +47,36 @@ def handle_list_import_status(request: flask.Request, ws_ns: str, ws_name: str)
...
+
+        if new_status == ImportStatus.Error:
+            imp.write_error(msg.get("error_message", "External service set this import to Error"))
```
davidangb: why do errors not require `current_status`?

davidangb (commented):
comments inline! Close to  approval but hoping for answers first.

app/status.py
helgridly: `update_status_exclusively` is meant to nix the problem of pubsub at-least-once

helgridly: `update_status_exclusively` is meant to nix the problem of pubsub at-least-once

helgridly: ugh, i'm already used to most handlers beginning with `update_status_exclusively`. Good call

app/status.py
helgridly: If someone's reporting an error, they're probably right. Quibbling about out-of-sequence messaging would leave the import stalled at a nonterminal status when we've already received information about why it broke, which doesn't feel good to me.

app/status.py
```diff
@@ -45,3 +47,39 @@ def handle_list_import_status(request: flask.Request, ws_ns: str, ws_name: str)
...
+        #   If the caller is setting to error, ignore current status and jump straight there.
+        #   If the import is already in a terminal status, the caller did something bad.
+        #   Otherwise update the status if the caller got the previous one correct.
+        if new_status == ImportStatus.Error:
+            imp.write_error(msg.get("error_message", "External service set this import to Error"))
+
+        elif imp.status in ImportStatus.terminal_statuses():
+            if new_status != imp.status:
+                raise exceptions.TerminalStatusChangeException(import_id, new_status, imp.status)
+            # else probably a pubsub double delivery, don't stress about it
```
helgridly: note: i also flipped the ordering here so that Error is always handled regardless

davidangb (approved):
:+1:

app/status.py
```diff
@@ -45,3 +47,41 @@ def handle_list_import_status(request: flask.Request, ws_ns: str, ws_name: str)
...
+        raise exceptions.BadJsonException(f"Missing current_status key from update status request for import {import_id}", audit_log = True)
+
+    update_successful = True
+    with db.session_ctx() as sess:
+        imp: model.Import = model.Import.get(import_id, sess)
+
+        # Only think about updating if the statuses are different.
+        if new_status != imp.status:
+            # Quick summary:
+            #   If the caller is setting to error, ignore current status and jump straight there.
```
ansingh7115, line 69: So - to be clear - if we are setting the status to Error, we do not need to pass pass in a "current_status"?

app/status.py
```diff
@@ -45,3 +47,41 @@ def handle_list_import_status(request: flask.Request, ws_ns: str, ws_name: str)
         import_statuses = [{"id": imprt.id, "status": imprt.status.name} for imprt in import_list]
 
         return flask.make_response((json.dumps(import_statuses), 200))
+
+
+def external_update_status(msg: Dict[str, str]) -> flask.Response:
+    """A trusted external service has told us to update the status for this import.
+    Change the status, but sanely.
+    It's possible that pub/sub might deliver this message more than once, so we need to account for that too."""
+    import_id = msg["import_id"]
```
ansingh7115: Instead of underscores, can these be camel case? Rawls object -> json conversions default to camel case.

ansingh7115 (approved):

app/status.py
helgridly: okay

app/status.py
helgridly: correct

PR #19, opened 2020-02-07
[noticket] gzip responses, plus grab bag of minor fixes

helgridly:
* gzip responses (> 500B)
* add `Content-Type: application/json` to outgoing responses
* move repeatedly defined fixtures to `conftest.py`
* dedupe repeated calls to `flask.make_response`

### DIFF
--- a/app/__init__.py
+++ b/app/__init__.py
@@ -2,10 +2,12 @@ import flask
 import flask_compress
 from app.server import routes, json_response
 
+compress = flask_compress.Compress()
+
 
 def create_app() -> flask.Flask:
     app = flask.Flask(__name__)
     app.response_class = json_response.JsonResponse
     app.register_blueprint(routes.routes)
-    flask_compress.Compress(app)
+    compress.init_app(app)
     return app
--- a/app/__init__.py
+++ b/app/__init__.py
@@ -1,10 +1,11 @@
 import flask
 import flask_compress
-from app.server import routes
+from app.server import routes, json_response
 
 
 def create_app() -> flask.Flask:
     app = flask.Flask(__name__)
+    app.response_class = json_response.JsonResponse
     app.register_blueprint(routes.routes)
     flask_compress.Compress(app)
     return app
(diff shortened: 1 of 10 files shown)


app/__init__.py
```diff
@@ -1,8 +1,13 @@
...
-from app.server import routes
+import flask_compress
+from app.server import routes, json_response
+
+compress = flask_compress.Compress()
 
 
 def create_app() -> flask.Flask:
     app = flask.Flask(__name__)
+    app.response_class = json_response.JsonResponse
```
helgridly, line 10: new response class that makes all responses have content-type app/json

app/__init__.py
```diff
@@ -1,8 +1,13 @@
...
     app.register_blueprint(routes.routes)
+    compress.init_app(app)
```
helgridly, line 12: gzip (see [here](https://github.com/colour-science/flask-compress#options) for under what conditions)

app/server/routes.py
```diff
@@ -13,21 +13,21 @@
 @httpify_excs
 def create_import(ws_ns, ws_name) -> flask.Response:
     """Accept an import request"""
-    return flask.make_response(new_import.handle(flask.request, ws_ns, ws_name))
+    return new_import.handle(flask.request, ws_ns, ws_name)
```
helgridly, line 16: handler functions return flask responses anyway so no need to so it twice (thanks @ansingh7115 )

app/tests/conftest.py
```diff
@@ -97,3 +97,23 @@ def pubsub_fake_env(monkeypatch) -> Iterator[None]:
 def fake_pfb() -> Iterator[IO]:
     with open("app/tests/empty.avro", 'rb') as out:
         yield out
+
```
helgridly, line 100: moved these fixtures from the top of multiple test classes to here

helgridly:
closing PR for now because we wanted the batchUpsert files to be gzipped, not API responses (and [GAE already gzips API responses for you anyway](https://cloud.google.com/appengine/docs/standard/go/how-requests-are-handled#response_compression))

PR #21, opened 2020-02-13
[AS-216] swagger using flask-restx

helgridly:
This uses [flask-restx](https://github.com/python-restx/flask-restx) to integrate Swagger. If you want to see it in action, DM me on Slack and I'll send you the link.

The auth is a bit funky. For our Docker-y services, the Apache proxy handles the oauth dance, so we don't have to think about it. Handling OAuth properly would require endpoints in import-service to do the dance ourselves, which isn't necessary in practice as this will hide behind Orchestration and not _need_ to do the dance. Swagger v3 supports auth-in-header but [restx is still only on v2](https://github.com/python-restx/flask-restx/issues/26). Thus we have to do a mildly gross thing and tell Swagger that we're authing using an API key, and expect the user to type the word `Bearer` before their token in the Authorize box. Given this Swagger will only be seen by internal developers this seemed like the right time/effort compromise.

### DIFF
--- a/app/__init__.py
+++ b/app/__init__.py
@@ -5,4 +5,5 @@ from app.server import routes
 def create_app() -> flask.Flask:
     app = flask.Flask(__name__)
     app.register_blueprint(routes.routes)
+    app.config["RESTX_MASK_SWAGGER"] = False  # disable X-Fields header in swagger
     return app
--- a/app/__init__.py
+++ b/app/__init__.py
@@ -1,9 +1,8 @@
 import flask
-from app.server import routes, json_response
+from app.server import routes
 
 
 def create_app() -> flask.Flask:
     app = flask.Flask(__name__)
-    app.response_class = json_response.JsonResponse
     app.register_blueprint(routes.routes)
     return app
--- a/app/health.py
+++ b/app/health.py
@@ -1,25 +1,37 @@
-import flask
-import json
-import logging
-from sqlalchemy.orm.exc import NoResultFound
-from typing import Dict
+from flask_restx import fields
 
-from app.auth import user_auth
 from app.db import db, model
-from app.db.model import ImportStatus
 from app.external import sam, rawls
-from app.util import exceptions
 
 
-def handle_health_check() -> flask.Response:
+class HealthResponse:
+    def __init__(self, db_health: bool, rawls_health: bool, sam_health: bool):
+        self.ok = all([db_health, rawls_health, sam])
+        self.subsystems = {
+            "db": db_health,
+            "rawls": rawls_health,
+            "sam": sam_health
+        }
 
+    @classmethod
+    def get_model(cls) -> model.ModelDefinition:
+        subsystem_fields = {
+            "db": fields.Boolean,
+            "rawls": fields.Boolean,
+            "sam": fields.Boolean
+        }
+        return {
+            "ok": fields.Boolean,
+            "subsystems": fields.Nested(subsystem_fields)
+        }
+
+
+def handle_health_check() -> HealthResponse:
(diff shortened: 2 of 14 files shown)


davidangb:
I like this approach better than #23 , at least until its constraints cause us problems when writing code. I think this approach makes the code more self-documenting, as well as generating swagger  documentation. I do recognize it's a lot of work to lean into restx.

MatthewBemis:
I think I like this one more? The alternative feels very tedious, verbose, and error-prone. This one feels like it's more intrusive but I suppose that's a reasonable cost for the benefits :)

app/__init__.py
```diff
@@ -1,9 +1,8 @@
 import flask
-from app.server import routes, json_response
+from app.server import routes
 
 
 def create_app() -> flask.Flask:
     app = flask.Flask(__name__)
-    app.response_class = json_response.JsonResponse
```
helgridly, line 7: this is handled magically for us by restx now

app/new_import.py
```diff
@@ -1,33 +1,12 @@
...
-    "filetype": {
-      "type": "string",
-      "enum": list(translate.FILETYPE_TRANSLATORS.keys())
-    }
-  },
-  "required": ["path", "filetype"]
-}
 
-
-schema_validator = jsonschema.Draft7Validator(NEW_IMPORT_SCHEMA)
```
helgridly, line 27: validation of incoming json is now handled by restx

app/server/routes.py
```diff
@@ -1,52 +1,87 @@
...
 routes = flask.Blueprint('import-service', __name__, '/')
 
+authorizations = {
+    'Bearer': {
+        "type": "apiKey",
+        "name": "Authorization",
+        "in": "header",
+        "description": "Use your GCP auth token, i.e. `gcloud auth print-access-token`. Required scopes are [openid, email, profile]. Write `Bearer <yourtoken>` in the box."
+    }
+}
```
helgridly, line 21: weird hack as described in the description

app/tests/test_new_import.py
```diff
@@ -24,9 +20,10 @@ def test_golden_path(client):
 
     # response contains the job ID, check it's actually in the database
     sess = db.get_session()
-    dbres = sess.query(Import).filter(Import.id == resp.get_data(as_text=True)).all()
+    id = resp.json["id"]
+    dbres = sess.query(Import).filter(Import.id == id).all()
     assert len(dbres) == 1
-    assert dbres[0].id == str(resp.get_data(as_text=True))
+    assert dbres[0].id == id
```
helgridly, line 26: tidied a bunch of tests up here

app/tests/test_health.py
```diff
@@ -0,0 +1,59 @@
+import flask.testing
```
helgridly, line 1: I had to rebase and add swagger for the new health check endpoint, so I wrote a test for it too.

app/health.py
```diff
@@ -1,29 +1,40 @@
...
-from app.db.model import ImportStatus
 from app.external import sam, rawls
-from app.util import exceptions
 
 
-def handle_health_check() -> flask.Response:
-
+class HealthResponse:
+    def __init__(self, db_health: bool, rawls_health: bool, sam_health: bool):
+        self.ok = all([db_health, rawls_health, sam])
```
MatthewBemis: this should be `self.ok = all([db_health, rawls_health, sam_health])`, right?

helgridly: good catch!

MatthewBemis (approved):

davidangb (commented):
retroactive  :+1:

PR #22, opened 2020-02-13
AS-141: Add status endpoint, support FiaB integration

MatthewBemis:


app/external/rawls.py
```diff
@@ -32,3 +32,13 @@ def check_workspace_iam_action(workspace_namespace: str, workspace_name: str, ac
         # just pass the error upwards
         logging.info(f"Got {resp.status_code} from Rawls for {workspace_namespace}/{workspace_name}: {resp.text}")
         raise ISvcException(resp.text, resp.status_code)
+
+
+def check_health() -> bool:
+    resp = requests.get(f"{os.environ.get('RAWLS_URL')}/status")
+
+    if resp.ok:
```
helgridly: `return resp.ok`

app/external/sam.py
```diff
@@ -87,3 +87,12 @@ def admin_get_pet_token(google_project: str, user_email: str) -> str:
     else:
         logging.debug(f"Got {resp.status_code} from Sam while trying to get pet key for {google_project}/{user_email}: {resp.text}")
         raise ISvcException(resp.text, resp.status_code)
+
+
+def check_health() -> bool:
+    resp = requests.get(f"{os.environ.get('SAM_URL')}/status")
+
+    if resp.ok:
```
helgridly: again

requirements.txt
```diff
@@ -24,3 +24,4 @@ gcsfs==0.6.0
 memunit==0.5.0
 psutil==5.6.7
 pyhumps==1.3.1
+gunicorn==20.0.4
```
helgridly: why does this need to be in requirements? if it's only used for the dockerfile i'd say install it there

helgridly (approved):
minor comments

app/external/rawls.py
MatthewBemis: yikes

PR #23, opened 2020-02-14
[EXPERIMENTAL] swagger using flasgger

helgridly:
This is one of two competing PRs that achieve the same thing. You should read them and compare! The other one is https://github.com/broadinstitute/import-service/pull/21.

This uses [flasgger](https://github.com/flasgger/flasgger) to integrate Swagger. I haven't fleshed out the documentation yet; this is a minimal integration that gets everything working.

As you can see, this is more explicit, and requires less reworking of application code -- at the expense of having to hand-roll docstrings everywhere. My definition of request and response bodies is lazy here, and could use a little more thought, but you get the gist.

### DIFF
--- a/app/__init__.py
+++ b/app/__init__.py
@@ -9,7 +9,17 @@ def create_app() -> flask.Flask:
     app.register_blueprint(routes.routes)
     app.config["SWAGGER"] = {
         'openapi': '3.0.2',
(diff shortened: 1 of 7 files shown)


helgridly:
closing since restx was preferred

PR #24, opened 2020-02-24
[AS-219] Pull from pubsub for FiaBs

helgridly:
This is part of a trifecta of PRs for getting import service working end to end on FiaB.

https://github.com/broadinstitute/import-service/pull/24
https://github.com/broadinstitute/rawls/pull/1184
https://github.com/broadinstitute/firecloud-develop/pull/2037

### DIFF
--- a/app/external/pubsub.py
+++ b/app/external/pubsub.py
@@ -1,6 +1,5 @@
 from google.cloud import pubsub_v1
-import google.cloud.pubsub_v1.types
-import os
+import os, contextlib
 from typing import Dict, List, Optional
 
 _publisher_client: Optional[pubsub_v1.PublisherClient] = None
@@ -14,13 +13,17 @@ def _get_publisher_client() -> pubsub_v1.PublisherClient:
     return _publisher_client
 
 
-def create_topic() -> None:
-    client = _get_publisher_client()
-    topic_path = client.topic_path(os.environ.get("PUBSUB_PROJECT"), os.environ.get("PUBSUB_TOPIC"))
-    try:
-        client.create_topic(topic_path)
-    except Exception:
-        pass  # it's fine if the topic already exists.
+def create_topic_and_sub() -> None:
+    pubclient = _get_publisher_client()
+    subclient = _get_subscriber_client()
+    topic_path = pubclient.topic_path(os.environ.get("PUBSUB_PROJECT"), os.environ.get("PUBSUB_TOPIC"))
+    sub_path = subclient.subscription_path(os.environ.get("PUBSUB_PROJECT"), os.environ.get("PUBSUB_SUBSCRIPTION"))
+
+    with contextlib.suppress(Exception): # it's fine if the topic already exists.
+        pubclient.create_topic(topic_path)
+
+    with contextlib.suppress(Exception): # it's fine if the subscription already exists.
+        subclient.create_subscription(sub_path, topic_path)
 
 
 def publish_self(data: Dict[str, str]) -> None:
--- a/app/external/pubsub.py
+++ b/app/external/pubsub.py
@@ -1,4 +1,5 @@
 from google.cloud import pubsub_v1
+import google.cloud.pubsub_v1.types
 import os
 from typing import Dict, List, Optional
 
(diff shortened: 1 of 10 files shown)


app/external/pubsub_pull.py
```diff
@@ -0,0 +1,37 @@
...
+@pubsubify_excs
+def process_one(attributes: dict) -> flask.Response:
+    return routes.pubsub_dispatch(attributes)
+
+
+def loop(app: flask.Flask):
+    """Main loop for pulling messages from PubSub when in PULL_MODE."""
+    while True:
+        try:
+            poll(app)
```
davidangb, line 20: introduce a `sleep` or some other way of not polling in a tight infinite loop. From what I can see this is hitting pubsub as fast as we possibly can (if we actually GET a message, we'll pause processing it)

davidangb (commented):
one comment inline, looks generally good. I''ll hold a thumb until this is officially ready for review.

app/external/pubsub_pull.py
helgridly: oh right, i forgot to do that -- thanks for the reminder!

davidangb (commented):
thank you for the sleep! Looks good, once this is ready for review and passes fiab tests I will thumb, or if there are changes I will re-review.

main.py
```diff
@@ -21,3 +21,13 @@
 
 
 app = create_app()
+
+# FiaB instances of this service live inside the Broad network and thus PubSub can't push notifications to the REST
+# handler. Setting PULL_PUBSUB will disable the REST handler and spin up a thread that pulls messages from PubSub instead.
+pull_pubsub = os.environ.get("PULL_PUBSUB", False)
+if pull_pubsub == "true" or pull_pubsub == "True":
```
davidangb: nit: could do `if pull_pubsub.lower() ==  "true"` to be lenient about config values

main.py
```diff
@@ -21,3 +21,13 @@
 
 
 app = create_app()
+
+# FiaB instances of this service live inside the Broad network and thus PubSub can't push notifications to the REST
+# handler. Setting PULL_PUBSUB will disable the REST handler and spin up a thread that pulls messages from PubSub instead.
```
davidangb: where does it disable the REST handler?

davidangb (commented):
question about the REST handler for pubsub ...

main.py
helgridly: excellent question! i couldn't tell if it was a good idea or not. what do you think? good: lock down unused endpoints. bad: might actually be useful for debugging, which is relevant since this is for a fiab

main.py
helgridly: wow i am dumb. thank you.

main.py
davidangb: I don't think it's necessary to turn off the unused endpoint - especially since this is in a fiab, not our prod instance. But  ... please update the comment!

davidangb (approved):
:+1: assuming you make the little change to update the comment about disabling pubsub push endpoint

app/external/pubsub.py
```diff
@@ -1,28 +1,62 @@
...
+def create_topic_and_sub() -> None:
+    pubclient = _get_publisher_client()
+    subclient = _get_subscriber_client()
+    topic_path = pubclient.topic_path(os.environ.get("PUBSUB_PROJECT"), os.environ.get("PUBSUB_TOPIC"))
+    sub_path = subclient.subscription_path(os.environ.get("PUBSUB_PROJECT"), os.environ.get("PUBSUB_SUBSCRIPTION"))
+
+    with contextlib.suppress(Exception): # it's fine if the topic already exists.
+        pubclient.create_topic(topic_path)
+
+    with contextlib.suppress(Exception): # it's fine if the subscription already exists.
```
ansingh7115, line 25: Would this suppress exceptions that aren't the topic/sub already existing?

helgridly: yes, but they'll be limited entirely to google-response-exceptions that we couldn't do anything about. google's pubsub library retries failures appropriately so we don't have to worry about handling transient things.

ansingh7115 (approved):

PR #26, opened 2020-03-03
AS-261 fix vuln that allowed you to download any file off the server filesystem

helgridly:
oopsie


davidangb (approved):
:+1: code looks fine, assuming you tested it does the job

PR #29, opened 2020-03-11
return error message in import status responses

helgridly:
now responses look like:

```
{
  "id" : "uuid-1234",
  "status" : "Pending",
  "error_message" : null
}
```

or if `status == Error`, then `error_message` will be a string

### DIFF
--- a/app/db/model.py
+++ b/app/db/model.py
@@ -10,7 +10,7 @@ from sqlalchemy_repr import RepresentableBase
 from app.db import DBSession
 
 from flask_restx import fields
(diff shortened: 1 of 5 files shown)


davidangb:
Is there a way to return
```
{
  "id" : "uuid-1234",
  "status" : "Pending"
}
```
if the `error_message` is None/null?

davidangb (approved):
:+1: approving; I'll tackle in a future PR the issue of omitting (or not) the null error message.

PR #30, opened 2020-03-18
AS-155: rename payloads

davidangb:
The get-import-status *response* payload now matches the payload currently returned by orch. This will allow orch to be purely passthrough for this endpoint, no additional logic; and should hopefully ease any end-to-end debugging.

We can't/shouldn't change the orch payloads, for backwards compatibility. Changing payloads of our public API would break the UIs.

I have *not* changed the create-import api,  since that would cause internal inconsistency inside import service.  For that api specifically, I'll munge both the request and response payloads inside orch.

### DIFF
--- a/app/db/model.py
+++ b/app/db/model.py
@@ -98,17 +98,29 @@ class ImportStatusResponse:
 
 # For backwards compatibility, new-import requests return a different response payload
 # than get-import-status requests.
-class NewImportResponse:
-    def __init__(self, url: str, jobId: str):
-        self.url = url
-        self.jobId = jobId
+class WorkspaceSpec:
+    def __init__(self, namespace: str, name: str):
+        self.namespace = namespace
+        self.name = name
 
     @classmethod
     def get_model(cls) -> ModelDefinition:
         return {
-            "url": fields.String,
-            "jobId": fields.String }
+            "namespace": fields.String,
+            "name": fields.String }
 
+class NewImportResponse:
+    def __init__(self, url: str, jobId: str, workspace: WorkspaceSpec):
+        self.url = url
+        self.jobId = jobId
+        self.workspace = workspace
+
+    @classmethod
+    def get_model(cls, api) -> ModelDefinition:
+        return {
+            "url": fields.String,
+            "jobId": fields.String,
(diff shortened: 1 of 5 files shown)


davidangb:
@helgridly what do you think, as lead dev for import service? @ansingh7115 most importantly do you see any problems with the import service <-> Rawls communication? I don't believe I've changed anything about the pubsub messages.

helgridly:
> Question: is it worth renaming "path" to "url" in the request payload to make that match orch?

No objection.

app/db/model.py
```diff
@@ -84,18 +84,43 @@ def from_string(cls, name: str):
...
     def get_model(cls) -> ModelDefinition:
         return {
-            "id": fields.String,
+            "jobId": fields.String,
             "status": fields.String,
-            "error_message": fields.String }
+            "message": fields.String }
 
+# For backwards compatibility, new-import requests return a different response payload
+# than get-import-status requests.
```
helgridly: Backwards compatibility with what?

app/db/model.py
```diff
@@ -84,18 +84,43 @@ def from_string(cls, name: str):
...
+class WorkspaceSpec:
```
helgridly: can we call this `model` or something that isn't `spec` -- that makes me think of tests

helgridly (approved):

ansingh7115 (approved):
Didn't see anything questionable!
EDIT: Actually, I did

app/db/model.py
```diff
@@ -84,18 +84,17 @@ def from_string(cls, name: str):
 # Note: this should really be a namedtuple but for https://github.com/noirbizarre/flask-restplus/issues/364
 # This is an easy fix in flask-restx if we decide to go this route.
 class ImportStatusResponse:
-    def __init__(self, id: str, status: str, error_message: Optional[str]):
-        self.id = id
```
ansingh7115, line 88: Actually, in rawls, we're looking for `id` and `status` when getting an import's status

ansingh7115: it throws away the error message

davidangb: good catch - broadinstitute/rawls#1188 should address that, yeah?

ansingh7115:
👍 Looks good now!

PR #31, opened 2020-03-19
deploy script should match app.yaml values [no ticket; risk: low]

davidangb:


helgridly (approved):
i long for the day (soon) when this script will die :)

PR #33, opened 2020-03-31
update ctmpl to be less inaccurate

helgridly:


app.yaml.ctmpl
```diff
@@ -1,23 +1,34 @@
 runtime: python37
-service: import-service
+service: default
```
helgridly, line 2: this is pretty important, fresh google projects must have an app engine project called "default" in them before deploying any differently-named ones. given that import service will live in its own project it has to be called default

app.yaml.ctmpl
```diff
@@ -1,23 +1,34 @@
...
+
+# see https://cloud.google.com/appengine/docs/standard/#instance_classes if we ever need to bump memory limits
+# instance_class: F4_HIGHMEM
+
+# in the ctmpl below, the token "vault" means secret/dsde/firecloud/<env>/import-service/
+# vault/foo.bar means vault read secret/dsde/firecloud/<env>/import-service/foo and look for the "bar" key
+# Note that you DON'T have to cover FiaB here, since on FiaB import-service is deployed using Docker and
+# environment variables in its docker-compose.
```
helgridly, line 10: more notes for you than anything else; feel free to nuke these once you're done

ansingh7115 (approved):

PR #34, opened 2020-05-08
AS-266: Deploy Import Service using Jenkins

ansingh7115:
Here are the Jenkins  jobs for deployment. This follows the example of apps like martha and bond that are independently deployable.

Manual non-prod deploy: https://fc-jenkins.dsp-techops.broadinstitute.org/job/import-service-manual-deploy/

Trigger non-prod deploy (occurs on merges to develop branch): https://fc-jenkins.dsp-techops.broadinstitute.org/job/import-service-deploy-trigger/

Manual production deploy: https://fcprod-jenkins.dsp-techops.broadinstitute.org/job/import-service-manual-deploy/


Dockerfile
```diff
@@ -1,12 +1,24 @@
 FROM python:3.7.4
```
ansingh7115: This Dockerfile is originally for standing up import service in FiaB. With this change, it also becomes the container from which to deploy import service. I don't know if it is a good idea to separate those two things into two different Dockerfiles

deploy.sh
```diff
@@ -0,0 +1,75 @@
...
+set +x
+if [ -z "$TARGET_ENV" ]; then
+    echo "TARGET_ENV argument not supplied; inferring from GIT_BRANCH '$GIT_BRANCH'."
+
+    if [ "$GIT_BRANCH" == "develop" ]; then
+        TARGET_ENV="dev"
+    elif [ "$GIT_BRANCH" == "alpha" ]; then
+        TARGET_ENV="alpha"
+    elif [ "$GIT_BRANCH" == "perf" ]; then
+        TARGET_ENV="perf"
```
ansingh7115, line 18: Currently, the [manual deploy job](https://fc-jenkins.dsp-techops.broadinstitute.org/job/import-service-manual-deploy/build?delay=0sec) only handles develop, alpha, and staging. 

Perf deploys are handled by the `deploy-all` -> `project-deploy` -> `DEPLOY` job. If we want import-service to be deployed on perf, we would need to either add import service to these jobs OR wait a bit and try to use the new GitHub Actions way of deploying perf

Dockerfile
davidangb: for deploy, could you use one of the prebuilt cloud sdk images? https://cloud.google.com/sdk/docs/downloads-docker

app.yaml.ctmpl
```diff
@@ -10,25 +22,29 @@ service: default
 # environment variables in its docker-compose.
 
 env_variables:
- CLOUD_SQL_CONNECTION_NAME: "mysql+pymysql://{vault/mysql/user.username}:{vault/mysql/user.password}@/isvc?unix_socket=/cloudsql/terra-importservice-{env}:us-central1:{vault/mysql/instance_details.instance_name}"
+ CLOUD_SQL_CONNECTION_NAME: "mysql+pymysql://{{$mysqlUsername}}:{{$mysqlPassword}}@/isvc?unix_socket=/cloudsql/terra-importservice-{{$env}}:us-central1:{{$mysqlInstanceName}}"
 
  PULL_PUBSUB: "False"
```
davidangb: this should be (pseudocode) `if fiab True, else False`

Edit: there may be other fiab issues in this template - ok to defer fiab work to a later PR.

app.yaml.ctmpl
```diff
@@ -10,25 +22,29 @@ service: default
...
  PUBSUB_TOPIC: "import-service-notify"
  PUBSUB_SUBSCRIPTION: "import-service-notify-push"
- PUBSUB_TOKEN: {vault/pubsub.secret_token}
- PUBSUB_ACCOUNT: "import-service@terra-importservice-{env}.iam.gserviceaccount.com"
+ PUBSUB_TOKEN: {{$pubSubToken}}
+ PUBSUB_ACCOUNT: "import-service@terra-importservice-{{$env}}.iam.gserviceaccount.com"
 
  # TODO: this needs to match the Terraform'd tfvars for the environment.
  # expect it to be importservice.firecloud.org for production and importservice.<env>.test.firecloud.org otherwise
  PUBSUB_AUDIENCE: "importservice.dev.test.firecloud.org"
```
davidangb: need to make this value env-specific

deploy.sh
```diff
@@ -0,0 +1,75 @@
...
+    echo "Unknown environment: $TARGET_ENV - must be one of [dev, alpha, perf, staging, prod]"
+    exit 1
+fi
+
+echo "Deploying branch '${GIT_BRANCH}' to ${ENVIRONMENT}"
+
+set -x
+
+GOOGLE_PROJECT=terra-importservice-${ENVIRONMENT}
+IMPORT_SERVICE_IMAGE=quay.io/broadinstitute/import-service:${GIT_BRANCH}
```
davidangb: when does this image get built/pushed? If you switch to using the prebuilt Cloud SDK image (see previous comment), not an issue.

deploy.sh
```diff
@@ -0,0 +1,75 @@
...
+docker pull $DSDE_TOOLBOX_DOCKER_IMG
+
+# render configs
+docker run -v $PWD:/app \
+  -e RUN_CONTEXT=live \
+  -e INPUT_PATH=/app \
+  -e OUT_PATH=/app \
+  -e VAULT_TOKEN=${VAULT_TOKEN} \
+  -e ENVIRONMENT=${ENVIRONMENT} \
+  -e ENV=dev \
```
davidangb: why does this have both `ENVIRONMENT` and `ENV`?

app.yaml.ctmpl
ansingh7115: Code comment from Hussein: 
` Note that you DON'T have to cover FiaB here, since on FiaB import-service is deployed using Docker and environment variables in its docker-compose.`

app.yaml.ctmpl
ansingh7115: ah - meant to ask you about this - is Hussein's code comment correct - are we using `importservice.firecloud.org` for production and `importservice.<env>.test.firecloud.org` for other envs?

deploy.sh
ansingh7115: ah! Was messing with something and didn't put it to rights afterwards - thanks!

app.yaml.ctmpl
davidangb: huh ... yeah, looking at the code: https://github.com/broadinstitute/import-service/blob/develop/main.py#L27 it seems as if this is not used. How about remove `PULL_PUBSUB` entirely from this file to avoid confusion?

app.yaml.ctmpl
davidangb: yup, the comment is correct. See the values in broadinstitute/terraform-ap-deployments#24 for reference!

app.yaml.ctmpl
ansingh7115: That works for me!

deploy.sh
MatthewBemis: it gets built/pushed by quay every time there's a merge to `develop`

Dockerfile
MatthewBemis: I think the idea of separating them makes more sense and would make this less confusing to maintain/understand. If I'm looking at this Dockerfile I'm not expecting it to also be used for deploying

app.yaml.ctmpl
```diff
@@ -10,25 +22,27 @@ service: default
...
+ SAM_URL: "https://sam.dsde-{{$env}}.broadinstitute.org"
 
- IMPORT_SVC_SA_EMAIL: "import-service@terra-importservice-{env}.iam.gserviceaccount.com"
- BATCH_UPSERT_BUCKET: "import-service-batchupsert-{env}"
+ IMPORT_SVC_SA_EMAIL: "import-service@terra-importservice-{{$env}}.iam.gserviceaccount.com"
+ BATCH_UPSERT_BUCKET: "import-service-batchupsert-{{$env}}"
 
- PUBSUB_PROJECT: "terra-importservice-{env}"
+ PUBSUB_PROJECT: "terra-importservice-{{$env}}"
  PUBSUB_TOPIC: "import-service-notify"
```
davidangb: this should be `"import-service-notify-{{$env}}"`

davidangb (approved):
lgtm!

MatthewBemis (approved):

PR #35, opened 2020-05-12
AS-326: prevent status updates from moving backwards [risk: low]

davidangb:
https://broadworkbench.atlassian.net/browse/AS-326

### DIFF
--- a/app/db/model.py
+++ b/app/db/model.py
@@ -47,12 +47,12 @@ else:
 
 @enum.unique
 class ImportStatus(enum.Enum):
-    Pending = enum.auto()  # import request received by the user but we haven't done anything with it yet
-    Translating = enum.auto()  # in the process of translating to rawls batchUpsert
-    ReadyForUpsert = enum.auto()  # batchUpsert file has been put in bucket and rawls has been notified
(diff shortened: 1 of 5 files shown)


app/db/model.py
```diff
@@ -47,12 +47,12 @@ def __init__(self, enum: Type[T]) -> None: ...
 
 @enum.unique
 class ImportStatus(enum.Enum):
-    Pending = enum.auto()  # import request received by the user but we haven't done anything with it yet
-    Translating = enum.auto()  # in the process of translating to rawls batchUpsert
-    ReadyForUpsert = enum.auto()  # batchUpsert file has been put in bucket and rawls has been notified
-    Upserting = enum.auto()  # rawls is actively working on importing the batchUpsert file
-    Error = enum.auto()  # something bad happened, check the error_message column for details
-    Done = enum.auto()  # success
```
ansingh7115, line 55: Python question - would setting these values to integers break the `from_string` function on line 73?

ansingh7115 (approved):

app/db/model.py
davidangb: @ansingh7115 `from_string` still works ... I have added a unit test to prove it!

ansingh7115 (approved):
Looks good!

PR #36, opened 2020-05-12
AS-326: (take two) ignore current when changing job status [risk: low]

davidangb:
#35 was incomplete in that it ignored the SQL check on current status. This updates that, including unit test.


ansingh7115 (approved):
Makes sense! Handles the case where import service processes status changes out of order

PR #37, opened 2020-05-12
add prod export host for BDC [no ticket; risk: low]

davidangb:
whitelists the `gen3-datastage-io-pfb-export.s3.amazonaws.com` location that BDC is currently using

Screenshot (11)


ansingh7115 (approved):
Approved - I think this is good for now - but perhaps we should move this to some sort of config later?

PR #43, opened 2020-06-04
add logging statements to import path [no ticket; risk: low]

davidangb:
adds a bunch of debugging logs to the code path that handles translation.

### DIFF
--- a/app/translate.py
+++ b/app/translate.py
@@ -1,5 +1,7 @@
 import flask
+import json
 import logging
+import traceback
 
 from app.auth import service_auth
 from app.auth.userinfo import UserInfo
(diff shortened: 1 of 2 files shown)


app/translate.py
```diff
@@ -37,10 +39,11 @@ def handle(msg: Dict[str, str]) -> ImportStatusResponse:
         # this import wasn't in pending. most likely this means that the pubsub message we got was delivered twice,
         # and some other GAE instance has picked it up and is happily processing it. happy translating, friendo!
         logging.info(f"Failed to update status exclusively for translating import {import_id}: expected Pending, got {import_details.status}. PubSub probably delivered this message twice.")
-        return flask.make_response(import_details.to_status_response())
+        return flask.make_response(json.dumps(import_details.to_status_response()))
```
davidangb, line 42: found via stackdriver; this was causing an error because `make_response` wanted a simpler type than `ImportStatusResponse`, so I'll turn it into a string.

ansingh7115 (approved):

PR #47, opened 2020-07-23
AS-456: error handling and debugging [risk: low]

davidangb:
fixes a few things I found while working on AS-456:
1. increase GAE hardware size, to try to make translation faster
2. use a new pubsub client for every message, so we don't try to reuse stale credentials and fail to auth
3. fix a serialization problem in an error case that made the error look worse than it was; this helps developers trying to debug problems
4. regularly log an "I'm still translating" message during translations, so we know the process is still active; this helps developers trying to debug problems

### DIFF
--- a/app.yaml.ctmpl
+++ b/app.yaml.ctmpl
@@ -28,8 +28,12 @@ basic_scaling:
 #
 # default instance_class for automatic_scaling is F1: 256MB / 600MHz
 # default instance_class for basic_scaling is B2: 512MB / 1.2GHz
+# other instance_class options:
+#   B4: 1024MB / 2.4GHz
+#   B4_1G: 2048MB / 2.4GHz
+#   B8: 2048MB / 4.8GHz
 #
-# instance_class: B2
+instance_class: B4
 
 # in the ctmpl below, the token "vault" means secret/dsde/firecloud/<env>/import-service/
 # vault/foo.bar means vault read secret/dsde/firecloud/<env>/import-service/foo and look for the "bar" key
--- a/app/external/pubsub.py
+++ b/app/external/pubsub.py
@@ -8,10 +8,11 @@ _subscriber_client: Optional[pubsub_v1.SubscriberClient] = None
 
 
 def _get_publisher_client() -> pubsub_v1.PublisherClient:
-    global _publisher_client
-    if _publisher_client is None:
-        _publisher_client = pubsub_v1.PublisherClient(credentials=service_auth.get_isvc_credential())
-    return _publisher_client
+    # global _publisher_client
+    # if _publisher_client is None:
+    #     _publisher_client = pubsub_v1.PublisherClient(credentials=service_auth.get_isvc_credential())
+    # return _publisher_client
+    return pubsub_v1.PublisherClient(credentials=service_auth.get_isvc_credential())
 
 
 def create_topic_and_sub() -> None:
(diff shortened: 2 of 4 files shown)


app/external/pubsub.py
```diff
@@ -44,10 +45,11 @@ def publish_rawls(data: Dict[str, str]) -> None:
...
 def _get_subscriber_client() -> pubsub_v1.SubscriberClient:
-    global _subscriber_client
-    if _subscriber_client is None:
-        _subscriber_client = pubsub_v1.SubscriberClient(credentials=service_auth.get_isvc_credential())
-    return _subscriber_client
+    # global _subscriber_client
+    # if _subscriber_client is None:
+    #     _subscriber_client = pubsub_v1.SubscriberClient(credentials=service_auth.get_isvc_credential())
+    # return _subscriber_client
+    return pubsub_v1.SubscriberClient(credentials=service_auth.get_isvc_credential())
```
davidangb, line 52: `_get_publisher_client` and `_get_subscriber_client` each previously created the pubsub client once and reused it infinitely for the life of the server instance. The clients include an auth token, which will expire. If the server instance has a long-running process the token can become invalid. Specifically, if Import Service needs to translate a very large PFB, the translation can take too long. Once the translation is done, Import Service is unable to publish the "please upsert" pubsub message to Rawls because the client's auth token expired.

app/translate.py
```diff
@@ -39,7 +40,7 @@ def handle(msg: Dict[str, str]) -> ImportStatusResponse:
         # this import wasn't in pending. most likely this means that the pubsub message we got was delivered twice,
         # and some other GAE instance has picked it up and is happily processing it. happy translating, friendo!
         logging.info(f"Failed to update status exclusively for translating import {import_id}: expected Pending, got {import_details.status}. PubSub probably delivered this message twice.")
-        return flask.make_response(json.dumps(import_details.to_status_response()))
+        return flask.make_response(f"Failed to update status exclusively for translating import {import_id}: expected Pending, got {import_details.status}. PubSub probably delivered this message twice.", 409)
```
davidangb, line 43: the `ImportStatusResponse` object was not dump-able via `json.dumps`, so this resulted in an error which was then caught in the `except Exception ... ` clause below, and logged to stackdriver as an error with ""Unexpcted error during translation". This error is misleading, since it's not actually unexpected nor should it be logged at error-level.

app/translate.py
```diff
@@ -87,12 +88,22 @@ def handle(msg: Dict[str, str]) -> ImportStatusResponse:
...
     for chunk in JSONEncoder(indent=0).iterencode(StreamArray(translated_gen)):
-        dest.write(chunk.encode())  # encodes as utf-8 by default
+        chunk_time = time()
+        num_chunks = num_chunks + 1
+        if (chunk_time - last_log_time >= 5):
+            elapsed = chunk_time - start_time
+            logging.info(f"still translating for import {import_id}: total time {elapsed}s, chunks processed {num_chunks}")
+            last_log_time = chunk_time
 
+        dest.write(chunk.encode())  # encodes as utf-8 by default
```
davidangb, line 106: these changes attempt to add some logging, every 5 seconds, to show that a translation process is still active.

app.yaml.ctmpl
```diff
@@ -28,8 +28,12 @@ basic_scaling:
 #
 # default instance_class for automatic_scaling is F1: 256MB / 600MHz
 # default instance_class for basic_scaling is B2: 512MB / 1.2GHz
+# other instance_class options:
+#   B4: 1024MB / 2.4GHz
+#   B4_1G: 2048MB / 2.4GHz
+#   B8: 2048MB / 4.8GHz
 #
-# instance_class: B2
+instance_class: B4
```
davidangb, line 36: increase GAE VM hardware, in an effort to make translations go a bit faster

MatthewBemis (approved):
My python and import-service knowledge is fairly limited, but thanks to your comments I feel pretty good giving this a 👍

PR #49, opened 2020-08-10
this should cause unit test failures [DO NOT MERGE]

davidangb:
testing CircleCI integration, this PR has an intentional unit test failure


davidangb:
circle is working as expected.

PR #50, opened 2020-10-06
AS-505: create all attrs in pfb: namespace [risk: low]

davidangb:
DO NOT MERGE - for review only. This will break the [cross-org BDC integration test](https://docs.google.com/document/d/1MXXjh6x9RfPwIqaesjq1k2V1r5KJ0Wt7P9P47ZIh4Zw/edit?usp=sharing). I have messages out to discuss how to fix that.

See https://docs.google.com/document/d/1B2QKTA2N9LsMrd0H1uRqtV2RbnunFNuSyirYUDkQzMI/edit?usp=sharing for full context.

This PR changes Import Service so that it creates all imported attributes in the `pfb:` namespace. It requires broadinstitute/rawls#1287 to succeed.

More accurately, it creates all imported attributes in a namespace equal to the "filetype" argument passed in to the import request. Currently, the only filetype that Import Service allows is "pfb" - but if in the future we support "avro" or "parquet" filetypes or something else, we'd create the attributes in those namespaces. We may also in the future add some other argument such as a "schemaName" or such.

_screenshot of PFB import, after this PR:_
Screenshot (4)

This PR will break an orchestration integration test. broadinstitute/firecloud-orchestration#805 is waiting and ready to fix that test.

### DIFF
--- a/app/tests/test_translate.py
+++ b/app/tests/test_translate.py
@@ -16,13 +16,13 @@ import pytest
 
 class StreamyNoOpTranslator(Translator):
     """Well-behaved no-op translator: does nothing, while streaming"""
-    def translate(self, file_like: IO) -> Iterator[Dict[str, Any]]:
+    def translate(self, file_like: IO, file_type: str) -> Iterator[Dict[str, Any]]:
         return ({line: line} for line in file_like)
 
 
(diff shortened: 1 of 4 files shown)


MatthewBemis (approved):
:+1:

PR #51, opened 2020-10-15
AS-505: also namespace the {entity_type}_name attr [risk: low]

davidangb:
follow-on to #50; I missed one of the attributes. All attributes should be namespaced.

### DIFF
--- a/app/translators/pfb_to_rawls.py
+++ b/app/translators/pfb_to_rawls.py
@@ -34,7 +34,9 @@ class PFBToRawls(Translator):
             if self.options['prefix-object-ids'] and key == 'object_id':
                 value = 'drs://' + value
             if key == 'name':
-                key = entity_type + '_name'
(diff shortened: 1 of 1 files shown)


app/translators/pfb_to_rawls.py
```diff
@@ -34,7 +34,9 @@ def make_op(key, value):
             if self.options['prefix-object-ids'] and key == 'object_id':
                 value = 'drs://' + value
             if key == 'name':
-                key = entity_type + '_name'
+                # with namespaces in place, why do we need the entity_type prefix here?
+                # we won't remove it now so as to not break any compatibility
+                key = file_type + ':' + entity_type + '_name'
```
davidangb, line 39: we need to have a conversation with Gen3/BDC and maybe HCA before removing the `entity_type_` prefix, just to be safe

MatthewBemis (approved):
makes sense 👍

PR #53, opened 2021-01-12
Add AppSec Github Trivy Action

dinvlad:
This change adds Trivy security scan for the Docker image, as discussed in the Workbench Cross-Team meeting.

The action will run on every PR and give you feedback if the Docker image contains critical OS vulnerabilities.

More info:
https://github.com/broadinstitute/dsp-appsec-trivy-action
https://github.com/broadinstitute/dsp-appsec-blessed-images

### DIFF
--- a/.github/workflows/trivy.yml
+++ b/.github/workflows/trivy.yml
@@ -1,0 +1,12 @@
+name: dsp-appsec-trivy
+on: [pull_request]
+
+jobs:
+  appsec-trivy:
+    # Parse Dockerfile and build, scan image if a "blessed" base image is not used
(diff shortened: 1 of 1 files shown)


davidangb:
@dinvlad fyi: import service is a Google App Engine app, and in production it does not use a docker image.  The Dockerfile in the repo is only used for testing in FiaBs. I'm not opposed to scanning it with Trivy ...  but does that add an extra repo for appsec to worry about unnecessarily?

dinvlad:
Makes sense, I was just unsure if it's GAE Standard or Flex! I'll go ahead and close it then, thanks for your input!

PR #54, opened 2021-01-13
move to appsec blessed base image [no ticket; risk: low]

davidangb:
The Dockerfile in this repo is only used inside of FiaBs. It is not used at runtime in production, since we deploy to GAE Standard instead.

For good measure, I'm updating the Dockerfile to use an appsec-blessed docker base image. As part of this move,  I had to refactor it to a multi-stage build, following the example at https://github.com/broadinstitute/dsp-appsec-blessed-images/tree/main/python.

I tested this by 1) verifying that Quay successfully built the image, and 2) manually running a swatomation-pipeline against the tag for this PR.

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -1,4 +1,4 @@
-FROM us.gcr.io/broad-dsp-gcr-public/base/python:distroless
+FROM us.gcr.io/broad-dsp-gcr-public/base/python:debian
 
 RUN apt-get update -y && \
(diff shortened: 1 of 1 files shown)


MatthewBemis (commented):
changes look good, did not test anything myself

MatthewBemis (approved):
whoops i hit "comment" not "approve"

ungwudik (approved):

PR #56, opened 2021-06-03
fixed dependency conflict from new pip versions

ssyms:
Updated psutil and memunit to avoid dependency conflict that arises in newer versions of pip. No new functionality, All pytests still pass.

### DIFF
--- a/requirements.txt
+++ b/requirements.txt
@@ -21,7 +21,7 @@ google-cloud-pubsub==1.1.0
 oauth2client==4.1.3
 pypfb==0.4.3
 gcsfs==0.6.0
-memunit==0.5.0
-psutil==5.6.7
+memunit==0.5.2
+psutil==5.7.0
 pyhumps==1.3.1
 flask-restx==0.1.1


sagehen03 (approved):

sagehen03:
I ran pytest.sh and mypy.sh with no errors and ran the service through docker and saw the swagger page.

davidangb (approved):
this worked for me … BUT only after I upgraded `pip` from 19.2.3 to 21.1.2

sagehen03:
There are some potentially latent issues when I build the docker image for example I get some red text: 

>  ERROR: After October 2020 you may experience errors when installing or updating packages. This is because pip will change the way that it resolves dependency conflicts.
> 
> We recommend you use --use-feature=2020-resolver to test your packages with the new resolver before it becomes the default.
> 
> google-api-core 1.30.0 requires google-auth<2.0dev,>=1.25.0, but you'll have google-auth 1.7.1 which is incompatible.
> google-cloud-core 1.6.0 requires google-auth<2.0dev,>=1.24.0, but you'll have google-auth 1.7.1 which is incompatible.
> pypfb 0.4.3 requires click~=7.0, but you'll have click 8.0.1 which is incompatible.
> WARNING: You are using pip version 20.2.3; however, version 21.1.2 is available.
> You should consider upgrading via the '/venv/bin/python3 -m pip install --upgrade pip' command.
>

PR #57, opened 2021-06-23
[AS-793] Update some dependencies, update AppEngine deployment target version

sagehen03:
Just updating the dependency versions as outlined https://broadworkbench.atlassian.net/browse/AS-793 resulted in a non-functioning deployment.  In fact, redeploying the same git commit that is currently working in dev resulted in a non-functional app in dev too.  Upgrading the python target version for app engine seems to have solved that first issue of a non-working app, but it led to others with the mypy linter.  The other changes in this PR were to make mypy happy.  I also changed circle ci python version to match the app engine target version.

### DIFF
--- a/app.yaml.ctmpl
+++ b/app.yaml.ctmpl
@@ -9,7 +9,7 @@
 {{- with $pubSubToken := $pubSub.Data.secret_token -}}
 
 
-runtime: python37
+runtime: python39
 service: default
 
 # set scaling algorithm for this service. basic_scaling parameters are:
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -6,7 +6,7 @@ version: 2
 jobs:
   build:
     docker:
-      - image: circleci/python:3.9.5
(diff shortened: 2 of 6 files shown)


app.yaml.ctmpl
```diff
@@ -9,7 +9,7 @@
 {{- with $pubSubToken := $pubSub.Data.secret_token -}}
 
 
-runtime: python37
+runtime: python39
```
davidangb, line 12: nit: could you also update:
* WALKTHROUGH.md to change "3.7" to "3.9"? I think that's the only place in the various readmes we list a specific version.
* Dockerfile to use 3.9.5 instead of 3.9.1, though I don't think this is a big deal

davidangb (approved):
minor comments inline, else :+1:

app.yaml.ctmpl
sagehen03: Updated.

ansingh7115 (approved):

PR #59, opened 2021-07-07
AS-460: Push docker images to GCR for use by FiaB

MatthewBemis:
https://github.com/broadinstitute/firecloud-develop/pull/2593 will be the follow-up to this to use the GCR images


minnie-kenny.gitconfig
```diff
@@ -0,0 +1,8 @@
+[secrets]
```
MatthewBemis, line 1: This repo didn't have minnie-kenny which seemed like an oversight? Regardless, I adapted a good amount of this from what Martha was doing and they have minnie-kenny too so figured I'd throw it in for symmetry reasons as well

.circleci/config.yml
```diff
@@ -47,4 +83,55 @@ jobs:
...
+
+      - setup_remote_docker:
+          docker_layer_caching: true
+
+      - run:
+          name: Build image
+          command: |
+            docker build -t "${GCR_IMAGE_NAME}:${CIRCLE_TAG:-${CIRCLE_BRANCH}}" -f Dockerfile .
+            if [[ ${CIRCLE_BRANCH} == "develop" ]]; then
+              docker tag ${GCR_IMAGE_NAME}:${CIRCLE_BRANCH} ${GCR_IMAGE_NAME}:latest
```
sagehen03, line 108: Not totally sure what's going on here, but it looks like CIRCLE_TAG might take precedence over CIRCLE_BRANCH.  If so, does the docker tag command on line 108 need to reference CIRCLE_TAG like line 106?

.circleci/config.yml
```diff
@@ -47,4 +83,55 @@ jobs:
...
+            # SA copied from secret/dsde/dsp-techops/common/dspci-wb-gcr-service-account.json
+            # Get the credentials used to push to GCR from CircleCI from the CircleCI environment variables
+            echo ${GCR_SA} > ${HOME}/gcr_auth_key_file.json
+            gcloud auth activate-service-account --key-file ${HOME}/gcr_auth_key_file.json
+            # Register `gcloud` as a Docker credential helper. Add `-q` to silence confirmation prompts
+            gcloud auth configure-docker ${GCR_REGISTRY_HOST} -q
+            docker push "${GCR_IMAGE_NAME}:${CIRCLE_TAG:-${CIRCLE_BRANCH}}"
+            if [[ ${CIRCLE_BRANCH} == "develop" ]]; then
+              docker push ${GCR_IMAGE_NAME}:latest
+            fi
```
sagehen03, line 122: It looks like in the case that we're on the develop branch we push to the docker repo twice--once with the longer name and once with the latest tag.  Do we incur the cost of uploading the image twice this way or does docker treat the second push as reference that it already knows about?  Just curious :)

sagehen03 (approved):
I haven't tested this and I'm only vaguely familiar with circle ci's job language so take my approval with a grain of salt.  I did have one question about the docker tag invocation that might be an edge case bug, but is more likely just my misunderstanding.

minnie-kenny.sh
```diff
@@ -0,0 +1,200 @@
+#!/bin/sh
+#   Use this script to ensure git-secrets are setup
+#   https://minnie-kenny.readthedocs.io/
```
davidangb, line 3: I'm fine with this PR as-is, am ok merging it … but in the same way that `run_secret_setup` clones/downloads git-secrets from https://github.com/awslabs/git-secrets.git, we should consider downloading minnie-kenny from https://github.com/broadinstitute/minnie-kenny.git instead of copying it.

davidangb (approved):
optional comment inline, otherwise :+1:

.circleci/config.yml
MatthewBemis: 106 should still be doing the tag for the CIRCLE_TAG case (with the build `-t` option), and 108 will only happen if CIRCLE_BRANCH is develop. I don't think CIRCLE_TAG could ever be "develop"

PR #62, opened 2021-07-12
support for async TSV uploads [AS-576]

davidangb:
See https://lucid.app/lucidchart/invitations/accept/inv_375c1c21-2df5-49de-9141-03b395aea2d8?viewport_loc=-310%2C105%2C2288%2C1190%2C1ImWrGZFNpE2 for a flowchart of the async import.

This PR adds support for a new "rawlsjson" filetype in import service, in addition to the only filetype we previously supported, "pfb."

The intention of the "rawlsjson" format is to allow Orchestration to support async TSV import. Orchestration already knows how to read a TSV and translate it into the json format Rawls needs … this is exactly what import-service does for PFBs. However, we need to leave the TSV-parsing code in Orchestration in order to support TSV downloads and to continue to support synchronous TSV uploads. So, we don't want to re-implement TSV parsing in import service. Instead, we'll let Orch create the json then pass that json to import service to handle the loading into Rawls.

Because this path requires no translation, all we do is move/rename the input file into the final location that import service/rawls expects. This bypasses the translation layer; no need to read the file, we just move it and let Rawls handle reading and validation. We do require that the file-to-be-imported is already in import-service's privileged bucket. This adds  GCS IAM protection: Orch has service account permissions to write to that bucket, but arbitrary callers cannot just load any file they want, unless they somehow get access to this bucket.

This PR is purely additive; deploying it should change nothing. Orch PR broadinstitute/firecloud-orchestration#861 is the one that actually exposes new functionality to users (at the API layer, not even in the UI yet).


app/translate.py
```diff
@@ -125,9 +138,13 @@ def validate_import_url(import_url: Optional[str], user_info: UserInfo) -> bool:
...
-    if any(actual_netloc.endswith(s) for s in VALID_NETLOCS):
+    if import_filetype == FILETYPE_NOTRANSLATION and actual_netloc == os.environ.get("BATCH_UPSERT_BUCKET"):
+        return True
+    elif import_filetype == "pfb" and any(actual_netloc.endswith(s) for s in VALID_NETLOCS):
         return True
     else:
-        logging.warning(f"Unrecognized netloc for PFB import: [{parsedurl.netloc}] from [{import_url}]")
-        raise exceptions.InvalidPathException(import_url, user_info, "PFB cannot be imported from this domain.")
+        logging.warning(f"Unrecognized netloc or bucket for import: [{parsedurl.netloc}] from [{import_url}]")
+        raise exceptions.InvalidPathException(import_url, user_info, "File cannot be imported from this URL.")
```
davidangb, line 152: there are probably other error-message strings that contain "PFB" - I haven't gone looking for them all.

requirements.txt
```diff
@@ -20,7 +20,7 @@ google-api-python-client==2.9.0
 google-cloud-pubsub==2.6.0
 oauth2client==4.1.3
 pypfb==0.4.3
-gcsfs==0.6.0
+gcsfs==2021.04.0
```
davidangb, line 23: `gcsfs` version 0.6.0 had errors using the `.mv()` command. Upgrading to `2021.04.0` fixed that. Note that `2021.04.0` is not the latest version; it is the last version before that library changed which Google APIs it uses under the covers, and it didn't seem risking that larger change for this POC.

app/translate.py
davidangb: I took a quick look and only found one other message, which I updated.

app/translate.py
```diff
@@ -105,7 +115,7 @@ def _stream_translate(import_id: str, source: IO, dest: IO, file_type: str, tran
 
         dest.write(chunk.encode())  # encodes as utf-8 by default
 
-def validate_import_url(import_url: Optional[str], user_info: UserInfo) -> bool:
+def validate_import_url(import_url: Optional[str], import_filetype: Optional[str], user_info: UserInfo) -> bool:
```
sagehen03, line 118: I'm not well versed in the type hinting library we use and it caused me some grief last time I was in this codebase so I might be biased, but should the import_url and import_filetype params be marked optional if we throw exceptions when they're None?  How about?
```suggestion
def validate_import_url(import_url: str, import_filetype: str, user_info: UserInfo) -> bool:
```

sagehen03 (approved):
One question about the type hints, but looks good!

app/translate.py
davidangb: hm, good question, I was just following the convention here. In the case where one of these values is actually None, I do like that we have full control over the log message and error sent to the user; that's nice.

That said, the first-pass validation actually happens in the model itself: https://github.com/broadinstitute/import-service/blob/1f617b7785fc2d1ac43575802a4426e27bd742df/app/server/routes.py#L33
so these error message should never be seen in practice

ansingh7115 (approved):
Looks good to me! Good tests.

PR #64, opened 2021-08-30
support upsert vs. update mode [AS-916]

davidangb:
AS-916

Import Service now accepts an optional isUpsert key in the request payload for creating an import job. If omitted, isUpsert defaults to true. The isUpsert value is persisted to the database as an attribute of the import job, and the value is sent along in the pubsub message that Import Service publishes to Rawls, to control the behavior of Rawls' writes. The corresponding Rawls PR is broadinstitute/rawls#1509.

This change is necessary to fully support asynchronous TSV import. Without this change, we could only support entity and membership TSVs, we could not support update TSVs.

TODO:
- [ ] this PR requires a new column in the the Import Service database table `imports`. Because we do not have a schema-migration system in place for Import Service (AS-184), I will manually create that column in the persistent environments once this PR is approved and ready to deploy.

### DIFF
--- a/requirements.txt
+++ b/requirements.txt
@@ -24,4 +24,4 @@ gcsfs==2021.04.0
 memunit==0.5.2
 psutil==5.7.0
 pyhumps==1.3.1
-flask-restx==0.1.1
+flask-restx==0.5.1
--- a/app/db/model.py
+++ b/app/db/model.py
@@ -7,6 +7,7 @@ from sqlalchemy import Column, DateTime, String
 from sqlalchemy.schema import Table
 from sqlalchemy.ext.declarative import declarative_base
 from sqlalchemy.orm import validates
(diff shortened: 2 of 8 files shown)


requirements.txt
```diff
@@ -24,4 +24,4 @@ gcsfs==2021.04.0
 memunit==0.5.2
 psutil==5.7.0
 pyhumps==1.3.1
-flask-restx==0.1.1
\ No newline at end of file
+flask-restx==0.5.1
```
davidangb, line 27: not strictly necessary as part of this PR, but I was debugging some flask-restx features and tried the upgrade, then left it in

MatthewBemis (approved):
lgtm

ungwudik (approved):
LGTM. Also want to acknowledge all the test coverage cases you added 👍

PR #66, opened 2021-09-04
possible bugfix for session handling

helgridly:
Hey folks. I still crib from import-service when I'm working on a Python project and I think I've noticed a mistake here. SQLAlchemy's `sessionmaker` is a session factory. Its intended usage is something like:

```
sessionmaker = sqlalchemy.orm.sessionmaker(bind=_db, ...blah blah...)

 # now get some sessions (imagine this is being called from other code)
session1 = sessionmaker()
session2 = sessionmaker()
```

Import service's code for `get_session()` is creating a sessionmaker every single time it's called. It's (probably) not going to cause any problems, but it only needs to be constructed once. This PR does that by creating the sessionmaker alongside the db engine singleton.

Obviously I haven't tested it on GAE etc. Hope y'all are well!

### DIFF
--- a/app/db/db.py
+++ b/app/db/db.py
@@ -17,6 +17,7 @@ _sessionmaker = None
 
 def get_session() -> DBSession:
     global _db
+    global _sessionmaker
 
     if _db is None:
         _db = sqlalchemy.create_engine(
--- a/app/db/db.py
+++ b/app/db/db.py
@@ -12,6 +12,7 @@ db_connection_name = os.environ.get("CLOUD_SQL_CONNECTION_NAME")
 
 # Store the db so it can be reused between GAE invocations.
 _db = None
+_sessionmaker = None
 
 
 def get_session() -> DBSession:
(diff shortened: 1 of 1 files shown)


davidangb (approved):
thank you!

davidangb (changes requested):
see #68, where we can get full tests running. Blocking merge until that other PR shows green …

davidangb:
@helgridly thanks again for this PR, and … good to hear from you!! See https://github.com/broadinstitute/import-service/pull/68/commits/52e89781364491ea7d85b4352133c86d2be8a35c … lots of tests are failing without this change. Do you want to incorporate that into this PR so it's your commit? I could also just merge #68.

helgridly:
Huh. With `_db` and `_sessionmaker` assigned in the same block, I don't understand why one would be `None` and not the other. But maybe there's some nuance to the GAE caching I'm missing.

In any case - go ahead and do whatever needs to be done to make this work for you. I'm not attached to having my name on the commit :)

davidangb:
I've merged #68; this PR is now obsolete. Thank you again! Feel free to close; I'll close it myself in a few days if you don't see this.

PR #71, opened 2021-10-22
AS-1004 revert to 3.9 base image

MatthewBemis:


Dockerfile
```diff
@@ -7,7 +7,7 @@ COPY requirements.txt /
 RUN /venv/bin/pip install -r /requirements.txt
 RUN /venv/bin/pip install gunicorn
 
-FROM us.gcr.io/broad-dsp-gcr-public/base/python:debian
+FROM us.gcr.io/broad-dsp-gcr-public/base/python:3.9-debian
```
davidangb, line 10: For good measure, could you try using this same image on line 1 for the `AS build` image too? Seems safest.

davidangb (approved):

sagehen03 (approved):

PR #72, opened 2021-12-01
AS-1035: parquet translation (but noop)

davidangb:
AS-1035

This adds a new filetype "tdrexport" to Import Service, alongside the pre-existing "pfb" and "rawlsjson".

In this PR, "tdrexport" is a no-op; it is hardcoded to produce zero rows for Rawls to import. The goal is to get an end-to-end spike working, even if it imports zero rows; later PRs/tickets will add rows to the working spike.


app/auth/service_auth.py
```diff
@@ -67,7 +67,7 @@ def _get_isvc_cached_creds() -> CachedCreds:
     """Use the cached creds if it still exists and we have at least 5 minutes until it expires."""
     if _cached_isvc_creds is not None and \
             _cached_isvc_creds.expiry > datetime.datetime.utcnow() + datetime.timedelta(minutes=5):
-        logging.info("using cached creds for import service SA")
+        logging.debug("using cached creds for import service SA")
```
davidangb, line 70: not necessary for this PR, but this cuts down on a lot of logging spam

app/translators/parquet_to_rawls.py
```diff
@@ -0,0 +1,17 @@
...
+    def __init__(self, options=None):
+        """Translator for Parquet files."""
+        if options is None:
+            options = {}
+        defaults = {}
+        self.options = {**defaults, **options}
+
+    def translate(self, file_like, file_type) -> Iterator[Dict[str, Any]]:
+        logging.info(f"executing a no-op ParquetToRawls translation for {file_type}: {file_like}")
+        return iter([])
```
davidangb, line 17: hardcoded to return an empty iterator - only valid for this end-to-end spike. Later PRs/Jira tickets will actually read the input file(s).

app/translate.py
```diff
@@ -24,7 +24,7 @@
 import requests.exceptions
 
 # these filetypes get stream-translated
-FILETYPE_TRANSLATORS = {"pfb": PFBToRawls}
+FILETYPE_TRANSLATORS = {"pfb": PFBToRawls, "tdrexport": ParquetToRawls}
```
davidangb, line 27: We may rename the filetype "tdrexport" in the future … once we see what the TDR export manifest actually looks like.

app/tests/test_parquet_to_rawls.py
```diff
@@ -0,0 +1,19 @@
...
+def test_noop_translate_parquet(fake_import_parquet):
+    """Proper translation of parquet files to rawls json."""
+    translator = ParquetToRawls()
+    result_iterator = translator.translate(fake_import_parquet, "tdrexport")
+    try:
+        item = next(result_iterator)
+    except StopIteration:
+        pass  # This is what should happen
+    else:
+        pytest.fail(f"Expected result_iterator to be empty, got {item}")
```
aherbst-broad: Would `pytest.raises()` work better here? ie., 
```suggestion
    with pytest.raises(StopIteration):
        item = next(result_iterator)
```

aherbst-broad (commented):
small comment about using pytest.raises

app/translate.py
```diff
@@ -8,7 +8,7 @@
 from app.db import db
 from app.db.model import *
 from app.external import pubsub
-from app.translators import Translator, PFBToRawls
+from app.translators import Translator, ParquetToRawls, PFBToRawls
```
jordansafer, line 11: I see we use ParquetToRawls and PFBToRawls. Is Rawls the official name of the format? Or is it defacto the name of the format at this point? I'm wondering if we're at risk of this becoming confusing if this shifts away from Rawls the service in a NextGenTerra situation-

app/translate.py
jordansafer: If we are using this to import other parquet files (for example, from Azure cloud), will those usecases also be TDR export cases? I'm trying to think if there's a usecase where we'd have parquet file imports that aren't TDR exports-

jordansafer (commented):
Looks good!! Just a couple of naming questions-

app/translate.py
ansingh7115: The format is Rawl's entity attribute format - see the request body example for this endpoint: https://rawls.dsde-dev.broadinstitute.org/#/entities/batch_update_entities

I imagine when we implement a new service, we wouldn't continue using this format, and it'll just die with rawls.

ansingh7115 (approved):

app/translate.py
jordansafer: I see, makes sense!

app/tests/test_parquet_to_rawls.py
davidangb: good call!

app/translate.py
davidangb: The predictions here are pretty fuzzy and in heavy discussion/brainstorming … I don't expect we'll have clarity on that question for a while. Are you suggesting renaming `ParquetToRawls` to something like `TDRToRawls`?

aherbst-broad (approved):

app/translate.py
jordansafer: Oh! I was thinking of changing the string "tdrexport" to "parquet" for filetype

jordansafer (approved):

app/translate.py
davidangb: the flow is definitely TDR-specific, as it will involve querying TDR to describe the snapshot being imported - that's why I didn't want to use a generic "parquet" as the filetype. Still, I'm totally open to renaming these as the implementation coalesces!

PR #75, opened 2021-12-03
add workspace_google_project to Import model [AS-1037]

davidangb:
part of AS-1037 but does not complete that ticket.

This PR adds a new `workspace_google_project` column/property to the `Import` object, and persists the workspace's googleProject (as returned by Rawls) for each new import.

This is a prereq for later work, in which we will need to get the user's pet inside that project and use that project for reading into GCS.

Because we do not yet have automatic DB schema migration for import service (AS-184), this PR will require manual handholding to alter the db schema in each env: 
```
ALTER TABLE imports ADD workspace_google_project varchar(30) NOT NULL default '' AFTER workspace_uuid;
```

### DIFF
--- a/app/external/rawls.py
+++ b/app/external/rawls.py
@@ -4,17 +4,24 @@ import os
 import requests
 
 from app.util.exceptions import ISvcException
+from dataclasses import dataclass
 from typing import Dict
 
 
-def get_workspace_uuid_and_project(workspace_namespace: str, workspace_name: str, bearer_token: str) -> Dict[str, str]:
+@dataclass
+class RawlsWorkspaceResponse:
+     workspace_id: str
+     google_project: str
+
+
+def get_workspace_uuid_and_project(workspace_namespace: str, workspace_name: str, bearer_token: str) -> RawlsWorkspaceResponse:
     resp = requests.get(
         f"{os.environ.get('RAWLS_URL')}/api/workspaces/{workspace_namespace}/{workspace_name}?fields=workspace.workspaceId,workspace.googleProject",
         headers={"Authorization": bearer_token})
 
     if resp.ok:
         jso = resp.json()
-        return {"uuid": jso["workspace"]["workspaceId"], "googleProject": jso["workspace"]["googleProject"]}
+        return RawlsWorkspaceResponse(workspace_id=jso["workspace"]["workspaceId"], google_project=jso["workspace"]["googleProject"])
     else:
         # just pass the error upwards
         logging.info(f"Got {resp.status_code} from Rawls for {workspace_namespace}/{workspace_name}: {resp.text}")
--- a/app/external/rawls.py
+++ b/app/external/rawls.py
@@ -4,15 +4,17 @@ import os
 import requests
 
 from app.util.exceptions import ISvcException
+from typing import Dict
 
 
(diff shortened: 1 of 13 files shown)


app/external/rawls.py
```diff
@@ -4,15 +4,17 @@
...
+def get_workspace_uuid_and_project(workspace_namespace: str, workspace_name: str, bearer_token: str) -> Dict[str, str]:
     resp = requests.get(
-        f"{os.environ.get('RAWLS_URL')}/api/workspaces/{workspace_namespace}/{workspace_name}?fields=workspace.workspaceId",
+        f"{os.environ.get('RAWLS_URL')}/api/workspaces/{workspace_namespace}/{workspace_name}?fields=workspace.workspaceId,workspace.googleProject",
         headers={"Authorization": bearer_token})
 
     if resp.ok:
-        return resp.json()["workspace"]["workspaceId"]
+        jso = resp.json()
+        return {"uuid": jso["workspace"]["workspaceId"], "googleProject": jso["workspace"]["googleProject"]}
```
aherbst-broad: Nitpick: This may be better represented as a dataclass rather than a freeform dict, with the added benefit of a bit more type safety (not sure if we run mypy in this repo) i.e,. 

```
@dataclass
class RawlsWorkspaceResponse:
     workspace_id: str
     google_project: str
```

aherbst-broad (approved):
Lgtm ; is there any kind of backfill that needs to happen to fill the new column for old imports?

davidangb:
@aherbst-broad good call on the dataclass! I've made those changes. 

I am not bothering with db backfill for existing imports. We only need the new column for newly-created TDR snapshot imports, and none of those exist yet (even if they did, they'd already be done or failed). We don't need to go back and act on those existing imports.

Take another look now that the dataclass changes are in?

aherbst-broad (approved):

app/external/rawls.py
```diff
@@ -4,15 +4,23 @@
...
+def get_workspace_uuid_and_project(workspace_namespace: str, workspace_name: str, bearer_token: str) -> RawlsWorkspaceResponse:
     resp = requests.get(
-        f"{os.environ.get('RAWLS_URL')}/api/workspaces/{workspace_namespace}/{workspace_name}?fields=workspace.workspaceId",
+        f"{os.environ.get('RAWLS_URL')}/api/workspaces/{workspace_namespace}/{workspace_name}?fields=workspace.workspaceId,workspace.googleProject",
         headers={"Authorization": bearer_token})
 
     if resp.ok:
-        return resp.json()["workspace"]["workspaceId"]
+        jso = resp.json()
+        return RawlsWorkspaceResponse(workspace_id=jso["workspace"]["workspaceId"], google_project=jso["workspace"]["googleProject"])
```
jordansafer, line 23: Continuing along the vein from @aherbst-broad, I think it would be good to use a serializer like Marshmallow or Pydantic, so we can statically check field accesses (both for field name and type)

jordansafer (approved):

app/external/rawls.py
davidangb: Using a serializer is a good idea! I think it's out of scope for this PR but we should queue up that work

jordansafer: I'll make a ticket!

PR #76, opened 2021-12-07
import TDR manifest [AS-1059]

davidangb:
* accept a TDR manifest file as the import URL
* read the manifest file from GCS as the importing user's pet SA
* parse the manifest file into json
* extract the table names, primary keys, and exported Parquet files for those tables from the manifest
* write Rawls data entities representing the tables


app/external/rawls_entity_model.py
```diff
@@ -0,0 +1,45 @@
...
+@dataclass
+class CreateAttributeEntityReferenceList(AttributeOperation):
+    attributeName: str
+    op: str = 'CreateAttributeEntityReferenceList'
+
+@dataclass
+class Entity:
+    name: str
+    entityType: str
+    operations: Sequence[AttributeOperation]
```
davidangb, line 45: we are not using all of these classes yet, but they will be necessary once we start parsing the Parquet files themselves, I'm looking forward

app/translate.py
```diff
@@ -1,30 +1,28 @@
...
-from typing import Dict, Optional, IO
-import os
-from json import JSONEncoder
-
-from urllib.parse import urlparse
-
-from gcsfs.core import GCSFileSystem
-import gcsfs.utils
-import requests.exceptions
-
```
davidangb, line 25: most of this is a result of my IDE's "organize imports"

app/external/rawls_entity_model.py
davidangb, line 52: we're not using all these classes in this PR, but they'll be necessary once we start parsing the actual Parquet exports, so I'm planning ahead a bit

app/external/rawls_entity_model.py
```diff
@@ -0,0 +1,45 @@
...
+@dataclass
+class AddUpdateAttribute(AttributeOperation):
+    attributeName: str
+    addUpdateAttribute: Any
+    op: str = 'AddUpdateAttribute'
+
+@dataclass
+class AddListMember(AttributeOperation):
+    attributeListName: str
+    newMember: Any
```
aherbst-broad: 🤔 Can this type be tightened up beyond `Any`?

app/external/tdr_manifest.py
```diff
@@ -0,0 +1,60 @@
...
+@dataclass
+class TDRTable:
+    name: str
+    primary_key: str
+    parquet_files: List[str] # TODO: list of urls instead? these are gs:// urls
+    reference_attrs: Dict[str, EntityReference] # column name -> reference
+
+
+class TDRManifestParser:
+    def __init__(self, jso: Dict[str, Any]):
```
aherbst-broad: It might be worthwhile to declare a JSON type to get some degree of type safety:
```
T = TypeVar('T')
JSON = Dict[str, T]
```

app/external/tdr_manifest.py
```diff
@@ -0,0 +1,60 @@
...
+        self._tables = self._parse()
+
+    def get_tables(self) -> List[TDRTable]:
+        return self._tables
+
+    def _parse(self) -> List[TDRTable]:
+        # the snapshot model: table names, primary keys, relationships
+        snapshot = self._jso['snapshot']
+        # the parquet export files
+        format = self._jso['format']['parquet']['location']  # pylint: disable=redefined-builtin
```
aherbst-broad: ⛏️ Have you considered renaming the var rather than disabling the pylint warning?

app/external/tdr_manifest.py
```diff
@@ -0,0 +1,60 @@
...
+                # extract primary key
+                tdr_pk = t['primaryKey']
+                if tdr_pk is None:
+                    pk = 'datarepo_row_id'
+                elif not isinstance(tdr_pk, list):
+                    pk = tdr_pk
+                elif isinstance(tdr_pk, list) and len(tdr_pk) == 1:
+                    pk = tdr_pk[0]
+                else:
+                    pk = 'datarepo_row_id'
```
aherbst-broad, line 52: 🤔 I'm trying to get my head around how we'd get here, seems like this is covered in L46?

app/translate.py
```diff
@@ -100,7 +105,10 @@ def handle(msg: Dict[str, str]) -> ImportStatusResponse:
 
 
 def _stream_translate(import_id: str, source: IO, dest: IO, file_type: str, translator: Translator) -> None:
-    translated_gen = translator.translate(source, file_type)  # doesn't actually translate, just returns a generator
+    # returns Iterator[Entity]
```
aherbst-broad: ⛏️ Remove this commented out code

app/translators/tdr_manifest_to_rawls.py
```diff
@@ -0,0 +1,49 @@
+import json
+import logging
+from typing import List  # pylint: disable=unused-import
```
aherbst-broad, line 3: Why not remove this and the other unused import below?

app/translators/tdr_manifest_to_rawls.py
```diff
@@ -0,0 +1,49 @@
...
+        self.options = {**defaults, **options}
+
+    def translate(self, file_like: IO, file_type: str) -> Iterator[Entity]:
+        logging.info(f'executing a TDRManifestToRawls translation for {file_type}: {file_like}')
+        # read and parse entire manifest file
+        jso = json.load(file_like)
+        tables = TDRManifestParser(jso).get_tables()
+
+        recs = []
+        ops = [] # type: List[AttributeOperation]
```
aherbst-broad: Is the `# type:` comment a hint for the static type checker? ```ops: List[AttributeOperation] = []``` may work as well

app/external/gcs.py
```diff
@@ -0,0 +1,15 @@
+from contextlib import contextmanager
+from typing import IO, Iterator
+
+from app.external import sam
+from gcsfs.core import GCSFileSystem
+
+# convenience function to read a GCS file as a user's pet SA.
+# this method is broken out from translate.py to make it easy to mock in unit tests.
```
aherbst-broad, line 8: Love the thinking about testability here.

aherbst-broad (approved):
Looking good, had a couple of questions.

app/external/tdr_manifest.py
davidangb: how would I do that? The json as provided by TDR has a key named "format": https://github.com/broadinstitute/import-service/blob/8bd78081888a196c651e41ebd82741ea163c9f20/app/tests/response_1638551384572.json#L785 … so don't I need to look for that key?

app/external/tdr_manifest.py
davidangb: this would include the case where tdr_pk is a list, but its length != 1

app/translators/tdr_manifest_to_rawls.py
davidangb: yes - this `# type:` comment is ready by mypy, and without this hint mypy complains. This is also why the imports of `List` and `AttributeOperation` are necessary - mypy needs the imports. It's an annoying situation:
* mypy complains without the `# type` hint and imports
* pylint complains about the otherwise-unused imports, unless I have `# pylint: disable`

app/external/tdr_manifest.py
```diff
@@ -0,0 +1,60 @@
...
+class TDRTable:
+    name: str
+    primary_key: str
+    parquet_files: List[str] # TODO: list of urls instead? these are gs:// urls
+    reference_attrs: Dict[str, EntityReference] # column name -> reference
+
+
+class TDRManifestParser:
+    def __init__(self, jso: Dict[str, Any]):
+        self._jso = jso
```
jordansafer: Are we planning to reuse self._jso outside of self._parse() in the future? If not, I'd prefer to pass it directly to the parse instead of making a class variable-

app/external/tdr_manifest.py
jordansafer: I think this is the case where the tdr primary key is a list, and length is greater than 1. I'm not familiar with the TDR format, so I'm not sure what it means for the table to have more than one primary key. Is there a good doc I could review on this?

app/external/rawls_entity_model.py
davidangb: check out the new `AttributeValue` alias, does this work?

davidangb:
@aherbst-broad take a quick look at the latest changes?

app/tests/response_1638551384572.json
```diff
@@ -0,0 +1,958 @@
...
+            "datatype": "string",
+            "array_of": false
+          },
+          {
+            "name": "descriptor",
+            "datatype": "string",
+            "array_of": false
+          }
+        ],
+        "primaryKey": null,
```
jordansafer, line 761: Would it make sense to test tables with different primaryKey formats?

app/tests/response_1638551384572.json
```diff
@@ -0,0 +1,958 @@
+{
```
jordansafer, line 1: This is super helpful!! It's really helpful to see what the output looks like here-

jordansafer (approved):
Is there a place where we have the spec for the metadata documented? 

Looks good!!

app/external/tdr_manifest.py
jordansafer: The warning is complaining about the variable name format, so the key to format is okay, but you could fix with something like:
```python
locations = self._jso['format']['parquet']['location']
```

davidangb: ohhhhhh duhhhhhhhhhhh

app/tests/response_1638551384572.json
davidangb: this is a good idea. I'd like to wait for TDR's export to be working as expected so we can see what those formats actually are … I am honestly not sure what all the possibilities are.

app/external/tdr_manifest.py
davidangb: good call!

jordansafer (approved):

PR #80, opened 2021-12-16
import rows from TDR-export Parquet files [AS-1037]

davidangb:
AS-1037

Inspect the TDR manifest to find all of the parquet files in a given export.

Read each of those files and parse them into Rawls entities. Save those entities into the upsert file for Rawls.

Need to create/identify follow-on Jira tickets for:
- [x] AS-1040: need a change in Rawls to allow new namespace(s), then move away from the `pfb:` namespace for attributes
- [x] AS-1072: handling of Parquet cells that are arrays - these will translate to RemoveAttribute+CreateAttributeValueList+[AddListMember] in Rawls json
- [x] AS-1073: don't query Sam for the exact same pet key for every parquet file
- [x] AS-1074: find memory limits. App Engine ran out of memory when attempting to import a snapshot with ~500k rows, but did ok on a snapshot with ~9k rows. Our driver projects are sized more like the latter, but still. We may want to increase app engine instance sizes, such as from B4 to B4_1G
- [x] AS-1075 re-enable the end-to-end `test_golden_path_tdr_manifest` test


app/tests/conftest.py
```diff
@@ -97,6 +97,11 @@ def fake_tdr_manifest() -> Iterator[IO]:
     with open("app/tests/response_1638551384572.json", 'rb') as out:
         yield out
 
+# TODO: delete this and the process-N.parquet files, so we don't have "real" data committed
```
jordansafer: Should this now be deleted?

app/tests/conftest.py
```diff
@@ -97,6 +97,11 @@ def fake_tdr_manifest() -> Iterator[IO]:
...
+# @pytest.fixture(scope="function")
+# def sample_tdr_parquet_file() -> IO:
+#     return open("app/tests/data/process/process-000000000000.parquet", 'rb')
+
```
sagehen03: Looks like we can delete this along with the process-N.parquet files.  Sorry, if I messed up by including that file in git.  I thought it was safe since it came out of dev.

app/translate.py
```diff
@@ -65,7 +65,7 @@ def handle(msg: Dict[str, str]) -> ImportStatusResponse:
 
             with filereader as pfb_file:
                 with gcs_project.open(dest_file, 'wb') as dest_upsert:
-                    _stream_translate(import_id, pfb_file, dest_upsert, import_details.filetype, translator = FILETYPE_TRANSLATORS[import_details.filetype]())
+                    _stream_translate(import_details, pfb_file, dest_upsert, import_details.filetype, translator = FILETYPE_TRANSLATORS[import_details.filetype]())
```
jordansafer: If we're passing the whole import_details object, we could skip on passing the filetype as a separate arg-

app/translators/tdr_manifest_to_rawls.py
```diff
@@ -1,49 +1,115 @@
...
+        self.filelocation = filelocation
+        self.file_nickname = os.path.split(filelocation)[1]
+
+    def translate(self) -> Iterator[Entity]:
+        """Converts a parquet file, represented as a url, to an iterator of Entity objects."""
+        logging.info(f'{self.import_details.id} attempting parquet translation of {self.file_nickname} from {self.filelocation} ...')
+        url = urlparse(self.filelocation)
+        bucket = url.netloc
+        path = url.path
+        # TODO: the call to gcs.open_file will get a new pet key each time. This is overly aggressive; we could probably
```
jordansafer, line 62: I vaguely remember conversation about a risk that the pet key could expire before a long import completes. Is this a realistic risk? What is our timeframe for uploads vs our petkey expiration time?

app/translators/tdr_manifest_to_rawls.py
```diff
@@ -1,49 +1,115 @@
...
+        pq_table: pyarrow.Table = pq.read_table(file_like)
+        df: pd.DataFrame = pq_table.to_pandas()
+        return self.translate_data_frame(df, pq_table.column_names)
+
+    def translate_data_frame(self, df: pd.DataFrame, column_names: List[str]) -> Iterator[Entity]:
+        """Convert a pandas dataframe - assumed from a Parquet file - to an iterator of Entity objects."""
+        logging.info(f'{self.import_details.id} expecting {len(df.index)} rows in {self.file_nickname} ...')
+        for _, row in df.iterrows():
+            ops = self.translate_parquet_row(row, column_names)
+            # TODO: better primary key detection/resilience?
```
jordansafer: Agreed/if we get incompatible data from TDR, how do we propagate that to the user?

app/tests/conftest.py
davidangb: I think it's probably safe but I lean pretty conservatively on this - I'd want to verify with the data owners first

app/translate.py
davidangb: good catch, thank you!

app/translators/tdr_manifest_to_rawls.py
davidangb: I'm going to create a follow-on story to optimize this. _Tokens_ expire but _keys_ should be valid for the length of the import, so we should reuse the key and reduce calls to Sam. I'm glad you asked this because I realize I had tokens and keys mixed up in this comment!

app/translators/tdr_manifest_to_rawls.py
davidangb: I've added logging for this, but as of now there's no good way to send warnings to the user for these async imports. We could either fail the entire import and deliver an error message, or allow the other rows to succeed but skip the messaging.

sagehen03 (approved):
You're a coding machine!  I'm probably not the best python reviewer, but I don't see any issues and I think since this is new functionality, and we can control how quickly clients start to use it that it seems safe to merge.

jordansafer (approved):

PR #81, opened 2021-12-17
update code comments [AS-1037]

davidangb:
I meant to commit this as part of #80 but I forgot to push it!

Just updates code comments and TODOs to make future development easier.

### DIFF
--- a/app.yaml.ctmpl
+++ b/app.yaml.ctmpl
@@ -55,7 +55,7 @@ env_variables:
  PUBSUB_TOKEN: {{$pubSubToken}}
  PUBSUB_ACCOUNT: "import-service@terra-importservice-{{$env}}.iam.gserviceaccount.com"
 
- # TODO: this needs to match the Terraform'd tfvars for the environment.
+ # this needs to match the Terraform'd tfvars for the environment.
  # expect it to be importservice.firecloud.org for production and importservice.<env>.test.firecloud.org otherwise
  PUBSUB_AUDIENCE:   {{if eq $env "prod"}}"importservice.firecloud.org"{{else}}"importservice.{{$env}}.test.firecloud.org"{{end}}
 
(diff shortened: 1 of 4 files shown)


app/external/tdr_manifest.py
```diff
@@ -59,7 +61,8 @@ def _parse(self, jso: JSON) -> List[TDRTable]:
                 else:
                     pk = 'datarepo_row_id'
 
-                # TODO: parse out reference attrs
+                # TODO AS-1036: from the "relationships" key in the manifest, save the valid reference attributes
+                # to reference_attrs. A reference is valid iif the "to" column is the primary key of the "to" table.
```
ungwudik: Teeny-tiny typo: meant to write `iff` instead?

ungwudik (approved):

app/external/tdr_manifest.py
davidangb: lol yes! fixed

app.yaml.ctmpl
```diff
@@ -55,7 +55,7 @@ env_variables:
  PUBSUB_TOKEN: {{$pubSubToken}}
  PUBSUB_ACCOUNT: "import-service@terra-importservice-{{$env}}.iam.gserviceaccount.com"
 
- # TODO: this needs to match the Terraform'd tfvars for the environment.
+ # this needs to match the Terraform'd tfvars for the environment.
```
sagehen03, line 58: What is involved in making this so?

sagehen03 (approved):
Left one question, but I'm sure it can be answered in the follow on work too if not now.

app.yaml.ctmpl
davidangb: it is already so! That's why I removed the TODO. I left the comment because if we ever change the yaml or the terraform, we'll need to keep them in sync

PR #82, opened 2021-12-17
support array attributes [AS-1072]

davidangb:
AS-1072

if the incoming data from TDR export is an array, create it as an array in Rawls.

Tested by deploying this branch to the dev env and seeing a successful import of an array value.

### DIFF
--- a/app/tests/test_tdr_manifest_to_rawls.py
+++ b/app/tests/test_tdr_manifest_to_rawls.py
@@ -2,11 +2,12 @@ import io
 from datetime import datetime
 from typing import Generator
 
+import numpy as np
 import pandas as pd
 import pyarrow as pa
 import pyarrow.parquet as pq
 from app.db.model import Import
-from app.external.rawls_entity_model import AddUpdateAttribute
+from app.external.rawls_entity_model import AddListMember, AddUpdateAttribute, CreateAttributeValueList, RemoveAttribute
 from app.external.tdr_manifest import TDRTable
 from app.translators.tdr_manifest_to_rawls import ParquetTranslator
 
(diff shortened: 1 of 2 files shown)


app/translators/tdr_manifest_to_rawls.py
```diff
@@ -118,15 +122,38 @@ def translate_parquet_attr(self, name: str, value: AttributeValue):
...
+            # most common case, results in AddUpdateAttribute
+            return [AddUpdateAttribute(usable_name, self.create_attribute_value(value))]
+        elif is_reference and is_array:
+            # TODO: AS-1038 detect/create references
+            # RemoveAttribute/CreateAttributeEntityReferenceList/AddListMember(s)
+            return []
+        elif is_array:
+            ops = [AddListMember(usable_name, self.create_attribute_value(v)) for v in value]
+            return [RemoveAttribute(usable_name), CreateAttributeValueList(usable_name), *ops]
+        else: # elif is_reference:
```
jordansafer, line 139: I'm not sure we need the is_reference cases here since we handle references in the create_attribute_value method

jordansafer (approved):

app/translators/tdr_manifest_to_rawls.py
davidangb: `create_attribute_value` will return an `EntityReference` object, but `translate_parquet_attr` will need to wrap that into a `AddUpdateAttribute(name, EntityReference(value))`, so I _think_ we need it in both … but this is definitely forward-looking code that will get filled in/overwritten by AS-1038

app/tests/test_tdr_manifest_to_rawls.py
```diff
@@ -95,23 +96,81 @@ def test_translate_parquet_file_with_missing_pk():
...
+    # note the differences in this data frame as compared to get_fake_parquet_translator()
+    file_like = io.BytesIO()
+    d = {'datarepo_row_id': ['a', 'b', 'c'], 'custompk': ['first', 'second', 'third'], 'arrayattr': [
+        ['Philip', 'Glass'],
+        ['Wolfgang', 'Amadeus', 'Mozart'],
+        ['Dmitri', 'Shostakovich']
+    ]}
+    df = pd.DataFrame(data=d)
+    table = pa.Table.from_pandas(df)
+    pq.write_table(table, file_like)
```
sagehen03: It looks like there are three places where we set up a file_like handle by writing a dict.  Can we extract that to a method that takes a dict and returns a file_like?

sagehen03 (approved):

PR #83, opened 2022-01-04
timestamp, source snapshotid, tdr namespace for all attributes [AS-1041]

davidangb:
depends on broadinstitute/rawls#1596

Three changes in this PR:
* for TDR import-by-copy, create all imported attributes inside the `tdr:` namespace. This avoids conflicts with reserved attribute names in Rawls ("name", "entityType", and "${entityType}_id". The last of those is the most common conflict.
* always add an "import:timestamp" attribute to each row, containing the timestamp of the import request
* always add an "import:snapshot_id" attribute to each row, containing the uuid of the snapshot from which we are importing


sagehen03 (approved):
Seems reasonable to me.

jordansafer (approved):

PR #84, opened 2022-01-04
re-enable golden path TDR import unit test [AS-1075]

davidangb:
AS-1075

This re-enables the "golden path" unit test for TDR import-by-copy. That test needed some extra mocking magic, which I've now implemented inside this PR.

### DIFF
--- a/app/tests/test_translate.py
+++ b/app/tests/test_translate.py
@@ -2,6 +2,7 @@ import io
 import os
 import unittest.mock as mock
 import urllib.error
+from contextlib import contextmanager
 from typing import IO, Iterator
 
 import gcsfs.retry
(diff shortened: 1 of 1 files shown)


app/tests/test_translate.py
```diff
@@ -14,6 +15,7 @@
 from app.tests import testutils
 from app.translators import Translator
 
+os.environ.setdefault("BATCH_UPSERT_BUCKET", "unittest-allowed-bucket")
```
davidangb, line 21: this is not technically necessary to re-enable the test. However, without this line present, you'll get test failures when running only the tests in this file. Tests will pass when running the entire suite of tests.

Seems that the tests inside this file relied on the `os.environ` setting in test_path_validation.py … when all tests were run, that setting allowed the tests inside here to pass.

sagehen03: Should we add a comment in the code as to why that line is necessary?

sagehen03 (approved):

jordansafer (approved):

PR #85, opened 2022-01-05
reuse pet keys for all Parquet files in a manifest [AS-1073]

davidangb:


sagehen03 (approved):
It's probably not worth worrying about, but it looks like in the event that the pet_key expires say during a long iteration over the parquet files there's no code to refresh the key and retry?  I don't know anything about pet keys so if you feel it's safe no need to add more code.

davidangb:
@sagehen03 keys don't expire (unless explicitly revoked) so I am not worried about this!

jordansafer (approved):

PR #86, opened 2022-01-07
[AS-1036][AS-1044] Order dependent tables for TDR import to Rawls

jordansafer:
Order dependent tables for importing TDR tables to Rawls. When we import TDR tables, the tables can depend on each other, and the tables that are depended on need to be imported first. This PR uses the relationship metadata provided in the TDR manifest file to order the tables for import

Changes:
- Use graphlib TopologicalSorter to order the tables based on dependent structure
- Add Pydantic typing and types for the manifest file
- Add reference_attributes fields for columns that refer to other tables


README.md
```diff
@@ -12,7 +12,7 @@ Create and activate the Python virtualenvironment:
 ```
 $ python3 -m venv venv
 $ source venv/bin/activate
-(venv) $ pip install --user -r requirements.txt
+(venv) $ python3 -m pip install -r requirements.txt
```
jordansafer, line 15: Got permission errors running this from a virtual env, so I updated this, also to make it clear which version of pip we are using (pip3/python3 pip).

app/external/tdr_manifest.py
```diff
@@ -1,23 +1,30 @@
...
+from app.external.tdr_model import Relationship, Table, TDRManifest
 
 
 @dataclass
 class TDRTable:
     name: str
     primary_key: str
     parquet_files: List[str]
-    reference_attrs: Dict[str, EntityReference] # column name -> reference
+    reference_attrs: Dict[str, str] # column name -> entity type (for referenced table)
```
jordansafer, line 20: Not passing entity reference because the value in an EntityReference is stored in the cell and is different for each row. It will be found in the TDR parquet file.

app/external/tdr_model.py
```diff
@@ -0,0 +1,88 @@
+## Pydantic model for TDR Manifest file, generated using https://github.com/koxudaxi/datamodel-code-generator
```
jordansafer, line 1: Added a Pydantic model to serialize JSON from. 2 questions here:
1. Does TDR have a model file they export (like Swagger)? This way we can merge in model changes with a version bump instead of having to own/update this file if the spec changes-
2. Is it possible to have extra fields or missing fields? Right now, Pydantic ignores extra fields, but fails on missing fields that aren't marked optional

davidangb: TDR does have a Python client that they publish. See this comment: https://docs.google.com/document/d/1LCDFMQZl1TTrJ8-vS2RMojkdledcKPFOaP19LjY9sig/edit?disco=AAAASbRNNyM, or see https://pypi.org/project/data-repo-client/. I had problems trying to use it a while back, BUT I was trying to use it to make requests, not just for the models.

We'd have to check with the Data Repo team about extra/missing fields, I don't know the answer to that!

app/external/tdr_manifest.py
```diff
@@ -28,44 +35,101 @@ def get_snapshot_id(self) -> str:
...
+        return pk
+    
+    # from the "relationships" key in the manifest, save the valid reference attributes to reference_attrs.
+    @staticmethod
+    def get_reference_attrs(relationships: List[Relationship], table_to_primary_key: Dict[str, str]) -> Dict[str, str]:
+        def relationship_to_reference(r: Relationship) -> Optional[Tuple[str, str]]:
+            # A reference is valid iff the "to" column is the primary key of the "to" table.
+            primary_key = table_to_primary_key[r.to.table]
+            if primary_key != r.to.column:
+                return None
```
davidangb, line 107: nit: could you log this case? I suspect we're going to run into user questions like "why aren't my entity references working??" If we had a log entry e.g. `TDR snapshot contains a relationship from foo.column to bar.column, but bar.column is not the primary key of bar`, we might be able to respond to those questions easier. In an ideal case, this log message would include the import job id and snapshot id, but if those are not easily available here, so be it

davidangb (commented):
one nit, one comment inline, looking good, thank you for this!

app/external/tdr_manifest.py
jordansafer: Definitely, I'll add a log

davidangb (approved):
thank you!

sagehen03 (approved):

PR #87, opened 2022-01-12
handle NaNs and better null/None handling [AJ-12]

davidangb:
AJ-12

When BigQuery/Parquet contained nulls in numeric columns (e.g. float), we sent `NaN` inside json to Rawls. This is invalid json and Rawls would fail the entire import. Now we send `null`.

When BigQuery/Parquet contained nulls in non-numeric columns we sent the string `"None"` inside json to Rawls. This was silly. Now we send `null`.


jordansafer (approved):
Looks great!

davidangb:
@jordansafer fyi I added the additional unit test, if you care to re-review

sagehen03 (approved):

jordansafer (approved):

jordansafer:
done- LGTM

PR #88, opened 2022-01-13
[AJ-108] add support for entity references

jordansafer:
https://broadworkbench.atlassian.net/browse/AJ-108

Added support for entity references from a TDR snapshot:
- check if the column is listed under "reference_attrs" to determine if it is a valid entity reference
- if yes and not an array, add the value as an entity reference with the entity_type provided from "reference_attrs"
- if yes and an array, create an entity reference list with a list of entity references
- added tests for entity reference and entity reference arrays


davidangb (approved):
lgtm, nice and concise!

sagehen03 (approved):

PR #90, opened 2022-02-15
[AJ-110] Sync permissions to TDR for snapshot import-by-copy

jordansafer:
https://broadworkbench.atlassian.net/browse/AJ-110

Syncing permissions to TDR by:
- Save the snapshot ID to the DB while translating the import to rawls-json
- After getting the completion notification from Rawls:
- - Get the pet service account token, and policy emails for the workspace from sam
- - Call TDR for each policy email and add it to the snapshot policy as a reader

Other minor changes:
- adding TDR endpoints
- fixing Pydantic model for the snapshot which was overly strict (errors surfaced during testing)
- add snapshot_id field to the db


app/status.py
```diff
@@ -86,7 +86,16 @@ def external_update_status(msg: Dict[str, str]) -> model.ImportStatusResponse:
...
                 current_status: ImportStatus = ImportStatus.from_string(msg["current_status"])
-                update_successful = model.Import.update_status_exclusively(import_id, imp.status, new_status, sess)
+
+                # we may need to sync permissions to tdr if this is a tdr snapshot
+                failed_sync = False
+                try:
+                    sync.sync_permissions_if_necessary(imp, new_status) 
+                except Exception as err:
+                    failed_sync = True
+                    imp.write_error(f"Failed to sync permissions for import {import_id}: {err}")
```
davidangb: A few tweaks here:
1. ~Inspect the `new_status` before attempting to sync permissions. If the `new_status` is not "Done", we shouldn't try to sync permissions. Otherwise, I believe this would also trigger when moving from `ReadyForUpsert` to `Upserting`, which we don't want … nor do we want to sync permissions if the import is moving from anything to `Error`.~
2.  The error message on failed sync is the message that will get bubbled up to the end user - can you make it more detailed about what happened? Assuming the `new_status` is `Done` before the sync attempt, could the error message be "All data imported successfully, but failed to synchronize permissions for import {import_id}: {err}" or similar?

Edit: ah, I see you check for "Done" in `sync_permissions_if_necessary()`, got it

app/translators/sync_permissions.py
```diff
@@ -0,0 +1,40 @@
...
+def sync_permissions_if_necessary(import_details: Import, import_status: ImportStatus):
+    """check if the status update is for a tdr snapshot sync that just completed, if yes, sync permissions"""
+    if import_details.status != ImportStatus.Done:
+        return # No sync required because import isn't done.
+
+    # if the import job doesn't come with a snapshot id, don't perform a sync
+    snapshot_id = import_details.snapshot_id
+    if snapshot_id is None:
+        # this should mean we aren't doing a tdr-export
+        return # no sync required since no snapshot present
```
davidangb, line 20: Your call: is it worth adding a check here for the condition of `if import_details.filetype == 'tdrexport' and snapshot_id is None:` and at least logging an error? By the time we get here, all `tdrexport`s should have a snapshot_id.

app/translators/sync_permissions.py
```diff
@@ -0,0 +1,40 @@
...
+    # call policy group emails and add them as readers to the snapshot
+    policy_group_emails: List[str] = get_policy_group_emails(import_details.workspace_uuid, pet_token)
+    for policy_group_email in policy_group_emails: 
+        tdr.add_snapshot_policy_member(snapshot_id, tdr.READER_POLICY_NAME, policy_group_email, pet_token)
+
+def get_policy_group_emails(workspace_id: str, bearer_token: str) -> List[str]:
+    """call sam to get all policies, and filter out policy group emails for groups that have read access"""
+    policies = list_policies_for_resource(WORKSPACE_RESOURCE, workspace_id, bearer_token)
+
+    reader_policies = filter(lambda policy: not set(policy.policy.roles).isdisjoint(set(READER_ROLES)), policies)
```
davidangb: I haven't tried this myself … could you use `issubset()` or `intersection()` here instead of `not … isdisjoint()`? It's a bit harder to think about the double negative.

app/translators/sync_permissions.py
```diff
@@ -0,0 +1,40 @@
...
+
+def sync_permissions(import_details: Import, snapshot_id: str):
+    """get a user's pet token, and use it to sync workspace readers to tdr to give them snapshot read access"""
+    # get the proper credentials to call as the user's pet service account
+    pet_token = sam.admin_get_pet_token(import_details.workspace_google_project, import_details.submitter)
+
+    # call policy group emails and add them as readers to the snapshot
+    policy_group_emails: List[str] = get_policy_group_emails(import_details.workspace_uuid, pet_token)
+    for policy_group_email in policy_group_emails: 
+        tdr.add_snapshot_policy_member(snapshot_id, tdr.READER_POLICY_NAME, policy_group_email, pet_token)
```
davidangb, line 35: here in `sync_permissions()` and also later in `get_policy_group_emails()`, I suggest adding logging statements for each external API call. In case of problem later we could hopefully track down where we hit problems.

davidangb (commented):
A few comments inline, looking really good!

I notice that `app/tests/test_sync_permissions.py` is committed but an empty file, was there supposed to be something there?

app/status.py
jordansafer: Makes sense! Adopted the more verbose import message-

app/translators/sync_permissions.py
jordansafer: Definitely worth adding a check and logging the error, it's a simple change and gives us something to look back on-

app/translators/sync_permissions.py
jordansafer: Switched to using intersection. I wish there was a hasintersection() which returned a bool, but no such equivalent for the positive case-

app/translators/sync_permissions.py
jordansafer: Added!

jordansafer:
> A few comments inline, looking really good!
> 
> I notice that `app/tests/test_sync_permissions.py` is committed but an empty file, was there supposed to be something there?

Oops! Great catch!! I missed adding/running these tests, and caught a bug just now with them.

app/external/sam.py
```diff
@@ -62,6 +75,66 @@ def get_user_action_on_resource(resource_type: str, resource_id: str, action: st
...
+                    "policy",
+                    "email"
+                ]
+            }
+        ]
+    }
+
+    resp = requests.get(
+        f"{os.environ.get('SAM_URL')}/api/resources/v2/{resource_type}/{resource_id}/policies",
+        headers={"Authorization": bearer_token}
```
davidangb, line 123: ah - this should be:
```
 headers={"Authorization": f"Bearer {bearer_token}"}
```

davidangb (approved):
lgtm! Did manual testing in the dev env show anything of note?

jordansafer:
> lgtm! Did manual testing in the dev env show anything of note?

Syncing succeeded for a steward of the TDR snapshot, and failed with a descriptive error in the UI for a reader of the TDR snapshot (missing permissions to add policy members)

sagehen03 (approved):
Looks like you haven't updated the oauth header to include Bearer as the prefix in the Authorization header value as David suggested?  Otherwise I didn't see any issues, but I didn't run the code, I'm not great at Python, and I don't know this particular domain very well :)  My understanding is we'll test some more once this is out in prod and the price of bugs is small for this nascent functionality.

jordansafer:
> Looks like you haven't updated the oauth header to include Bearer as the prefix in the Authorization header value as David suggested? Otherwise I didn't see any issues, but I didn't run the code, I'm not great at Python, and I don't know this particular domain very well :) My understanding is we'll test some more once this is out in prod and the price of bugs is small for this nascent functionality.

Yup! I put the Bearer prefix in a function that serves up the auth_token, since the token is used in a couple places, so it's a little less brittle that way. And sounds good, thank you!

PR #92, opened 2022-03-01
additional logging around permission-sync [AJ-111]

davidangb:
In my work to write integration tests (broadinstitute/firecloud-orchestration#916), I'm seeing some test failures that are difficult to debug. Here, I add a couple more logging statements to try to track down what's happening inside Import Service to cause failed imports during the permission-sync step.

### DIFF
--- a/app/status.py
+++ b/app/status.py
@@ -1,5 +1,6 @@
 import flask
 import logging
+import traceback
 from sqlalchemy.orm.exc import NoResultFound
 from typing import Dict, List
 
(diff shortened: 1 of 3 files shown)


app/status.py
```diff
@@ -93,6 +94,7 @@ def external_update_status(msg: Dict[str, str]) -> model.ImportStatusResponse:
                     sync.sync_permissions_if_necessary(imp, new_status)
                 except Exception as err:
                     failed_sync = True
+                    logging.error(f"Error during permission syncing for import {import_id}: {traceback.format_exc()}")
```
jordansafer, line 97: Seems weird to me that we don't pass err as an argument to `traceback.format_exc()`, but I just tried it, and it works! strange library...

jordansafer (approved):

sagehen03 (approved):

jordansafer (approved):

PR #93, opened 2022-03-02
[AJ-320] don't add tdr prefix for non-problematic rows

jordansafer:
https://broadworkbench.atlassian.net/browse/AJ-320

Only put tdr prefix where it's needed, see here for cases:
https://docs.google.com/document/d/1_dEbPtgF7eeYUNRFK6CDUqeGWtiE9ISeTlfjSEtl-FA/edit#

### DIFF
--- a/app/tests/test_tdr_manifest_to_rawls.py
+++ b/app/tests/test_tdr_manifest_to_rawls.py
@@ -7,7 +7,6 @@ import numpy as np
 import pandas as pd
 import pyarrow as pa
 import pyarrow.parquet as pq
-from sqlalchemy import null
 from app.db.model import Import
 from app.external.rawls_entity_model import AddListMember, AddUpdateAttribute, AttributeOperation, \
     CreateAttributeEntityReferenceList, CreateAttributeValueList, Entity, EntityReference, RemoveAttribute
@@ -293,7 +292,6 @@ def test_namespace_added_where_required():
 
     # import should always be namespaced
     assert translator.add_namespace_if_required('import:fake_timestamp') == 'import:fake_timestamp'
-
 
 def test_if_namespace_prefix_will_be_added():
     # no additional prefix required if a prefix is already present to make this valid
--- a/app/tests/test_tdr_manifest_to_rawls.py
+++ b/app/tests/test_tdr_manifest_to_rawls.py
@@ -40,13 +40,13 @@ def test_translate_data_frame():
     assert len(entities) == 3
     assert entities[0].name == 'a'
     assert entities[0].entityType == 'unittest'
-    assert entities[0].operations == [AddUpdateAttribute('tdr:datarepo_row_id', 'a'), AddUpdateAttribute('tdr:one', 1), AddUpdateAttribute('tdr:two', 'foo'), _import_sourceid(random_snapshot_id), _import_timestamp(now)]
+    assert entities[0].operations == [AddUpdateAttribute('datarepo_row_id', 'a'), AddUpdateAttribute('one', 1), AddUpdateAttribute('two', 'foo'), _import_sourceid(random_snapshot_id), _import_timestamp(now)]
     assert entities[1].name == 'b'
     assert entities[1].entityType == 'unittest'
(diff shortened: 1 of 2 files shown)


app/translators/tdr_manifest_to_rawls.py
```diff
@@ -165,3 +165,11 @@ def create_attribute_value(self, value, is_reference: bool = False, reference_ta
...
                 # and str() the rest of them.
                 return str(value)
+
+    # only add TDR namespace if needed. See this doc:
+    # https://docs.google.com/document/d/1_dEbPtgF7eeYUNRFK6CDUqeGWtiE9ISeTlfjSEtl-FA
+    @staticmethod
+    def prefix_required(name: str, namespace: str, table_name: str, primary_key: str) -> bool:
+        return namespace != 'tdr' \
+            or name == 'name' \
+            or (name.endswith('_id') and name[:-3] == table_name and name != primary_key)
```
davidangb: A few comments on this logic:
1. the attribute name "entityType" also needs to be prefixed. This is discussed in the Background section of the linked doc, but not in the examples - sorry about that!
2. comparison logic here needs to be case-insensitive. "Name" is disallowed equally to "name"; tablename+"_iD" is disallowed equally to tablename+"_id"

davidangb: And, some optional thoughts - I won't block the PR on this: it feels kinda dirty to be setting the "tdr" namespace on every call to `translate_parquet_attr`, but then dropping it by checking specifically for that value in here. What if we did something like:
* remove the `namespace` argument from `translate_parquet_attr()`
* modify the two calls in `translate_parquet_row()` to send name arguments of "import:timestamp" and "import: snapshot_id" to `translate_parquet_attr()`
* add a new method e.g. 
```python
    @staticmethod
    def rename_if_required(name: str, table_name: str, primary_key: str) -> str:
        if prefix_required(name, table_name, primary_key):
            f'tdr:{name}'
        else:
            name

```
and then modify `translate_parquet_attr`:
```python
usable_name = ParquetTranslator.rename_if_required(name, self.table.name, self.table.primary_key)
```

this still hardcodes "tdr" in one place, but it only comes into play for those infrequent attributes that require prefixes.

app/tests/test_tdr_manifest_to_rawls.py
```diff
@@ -264,10 +270,29 @@ def assert_attr_value(ops: Sequence[AttributeOperation], attrname: str, expected
...
+
+    # prefix is required if it's not the primary key but is tableName_id
+    assert ParquetTranslator.prefix_required('sample_id', tdr_namespace, 'sample', 'notsample_id')
+
+    # otherwise no prefix!
+    assert not ParquetTranslator.prefix_required('sample_id', tdr_namespace, 'sample', 'sample_id')
+    assert not ParquetTranslator.prefix_required('aliquot_id', tdr_namespace, 'sample', 'sample_id')
+    assert not ParquetTranslator.prefix_required('bam', tdr_namespace, 'sample', 'sample_id')
+    assert not ParquetTranslator.prefix_required('aliquot_id', tdr_namespace, 'sample', 'aliquot_id')
+    assert not ParquetTranslator.prefix_required('sample_id_id', tdr_namespace, 'sample', 'notsample_id')
```
davidangb: see my comments on the runtime `prefix_required()` - probably want a few more test cases in here. I'd also suggest adding one or more test cases where the primary key is null/None, since we tend to see that a lot.

davidangb (commented):
thanks for cranking this out so quick! See comments inline …

sagehen03 (approved):
modulo David's comments on test cases covering null pk's

app/tests/test_tdr_manifest_to_rawls.py
jordansafer: Ah! Added some testing for cases & None-ness

app/translators/tdr_manifest_to_rawls.py
jordansafer: Agreed with all of these things! Minor-ly didn't make the new method static because it saves some argument forwarding (passing self.table.name to rename_if_require just to pass it to prefix_if_required), but otherwise the same!

Fixes for the first 2 and optional cleanup in the updated PR-

davidangb (approved):
lgtm, thank you!

PR #96, opened 2022-03-30
AJ-13: stop import and don't download file if it's too big

sagehen03:
I fought a good bit with the mocking.  It seems like I should not have to pass file system mock to `gcs.open_file` but I couldn't get it to work without that.  I will test end to end manually in dev once I deploy to dev.


app/external/gcs.py
```diff
@@ -6,20 +6,28 @@
...
-# this method is broken out from translate.py to make it easy to mock in unit tests.
+from app.util.exceptions import FileTooBigToDownlod
 
+
+# convenience function to read a GCS file as a user's pet SA
+# this method is broken out from translate.py to make it easy to mock in unit tests
 @contextmanager
-def open_file(project: str, bucket: str, path: str, submitter: str, auth_key: Dict[str, Any] = None) -> Iterator[IO]:
+def open_file(project: str, bucket: str, path: str, submitter: str, auth_key: Dict[str, Any] = None,
+              file_limit_bytes: int = 2147483648, gcsfs: GCSFileSystem = None) -> Iterator[IO]:
```
sagehen03: I'm not thrilled about adding the GCSFileSystem param here, but I could not get the tests to properly patch GCSFileSystem with a mock.  Would love it if we could remove this and still have a unit test validating the expected behavior.

app/external/gcs.py
```diff
@@ -6,20 +6,28 @@
...
         logging.debug(f'retrieving auth key from Sam to read {path}')
         auth_key = sam.admin_get_pet_key(project, submitter)
 
     try:
-        manifest_fs = GCSFileSystem(project=project, token=auth_key)
-        with manifest_fs.open(f"{bucket}{path}") as response:
+        fs = GCSFileSystem(project=project, token=auth_key) if (gcsfs is None) else gcsfs
+        # du() would seem to be a more straightforward option, but it requires permissions that TDR doesn't grant,
+        # so we use info, if size metadata is not present don't download
+        if fs.info(f"{bucket}{path}").get('size', file_limit_bytes) >= file_limit_bytes:
```
sagehen03, line 29: I feel some ambivalence about throwing an exception here if the file size isn't present.

app/util/exceptions.py
```diff
@@ -19,7 +19,7 @@ def __init__(self, message: str, http_status: int = 500, imports: Optional[List[
         self.retry_pubsub = retry_pubsub
         self.audit_logs = audit_logs if audit_logs else []
         self.imports = imports if imports else []
-
+        super().__init__(message)
```
sagehen03, line 22: Before this change the underlying error message was not being reported.  I'm not sure if that was on purpose.

app/util/http.py
```diff
@@ -1,11 +1,17 @@
...
 
+from app.util.exceptions import FileTooBigToDownlod
+
 
 @contextmanager
-def http_as_filelike(url: str) -> Iterator[IO]:
+def http_as_filelike(url: str, file_limit_bytes: int = 2147483648) -> Iterator[IO]:
     """Open a file over HTTP and return it as a file-like object."""
+    http_response = requests.head(url)
+    if int(http_response.headers.get('content-length', file_limit_bytes)) >= file_limit_bytes:
```
sagehen03, line 14: similar to gcs, if the server won't tell us about the file before we try to download it, then don't download it.

davidangb: This is called for PFB imports. Do you know if this will cause any problems for BioData Catalyst, LungMAP, etc. who are currently using PFBs? Generally these are signed URLs in S3 or GCS, do you know if those reliably report a content-length here?

davidangb (commented):
one open question inline, rest of the code looks good!

app/util/http.py
sagehen03: As per our standup conversation, I'm going to test some urls for PFB import and attempt to verify.

sagehen03: I tested BDC and LungMAP and they both worked.  I also tested a file in excess of the limit and it failed as expected.

davidangb (approved):

jordansafer (approved):

PR #97, opened 2022-03-30
[AJ-365] prefix primary_key when adding it as an attribute

jordansafer:
The issue was our handling of a table with entityType 'sample' and primaryKey 'sample_id'. First, we would look up the value in the 'sample_id' column and use it as our primaryKey. Next we would add the 'sample_id' attribute: AddUpdateAttribute('sample_id',...). This fix prefixes the attribute with 'tdr:'

Note: If the primaryKey is different from the entityType (which we pull from the TDR table name), the behavior can be confusing. For example, if the entityType is 'sample' and the primaryKey is 'aliquot_id', in Terra we will have a primaryKey named 'sample_id', with the values taken from 'aliquot_id'. We will also have an attribute column 'aliquot_id' with the same values. We are trusting that 'table name' is a source of truth, and will be set correctly. In the future, if the table name was set incorrectly in TDR, a user could use that 'change entityType' functionality when it is complete to make changes.

Tested in dev by exporting a new test snapshot and confirming the successful upsert to rawls

EDIT/Update:
In the event a tableName matches a primaryKey, for example in the case where the tableName is 'sample' and the 'primaryKey' is 'sample_id', we will only add the primaryKey as the primaryKey, and we won't add it as an Attribute. If the primaryKey went under a different name, like 'aliquot_id', we would still add both the attribute (which would appear under aliquot_id as an Attribute column) and the primaryKey (which would appear under sample_id in the primaryKey column).


requirements.txt
```diff
@@ -3,6 +3,7 @@ types-requests==2.26.1
 setuptools==59.5.0
 wheel==0.37.0
 Flask==2.0.2
+werkzeug==2.0.3 # later version incompatible with Flask, newer Flask incompatible with pypfb
```
jordansafer, line 6: import-service build started failing for me (latest version, with no code changes) due to a breaking update in this dependency. Added this line to fix the dependency version.

app/tests/test_tdr_manifest_to_rawls.py
```diff
@@ -279,6 +270,35 @@ def assert_attr_value(ops: Sequence[AttributeOperation], attrname: str, expected
...
+
+    # 1000 Genomes public data. This Parquet file contains one row. That row
+    # contains nulls in float and boolean columns
+    with open('app/tests/resources/short_sample_table.parquet', 'rb') as file_like:
+        entities = list(translator.translate_parquet_file_to_entities(file_like))
+        assert len(entities) == 1
+        e: Entity = entities[0]
+        print(e)
+        print(e.operations)
+        print(e.name)
```
davidangb: nit: delete `print` statements

davidangb (commented):
:+1: on syntax, with one nit inline. I think this needs some PO input on behavior, though - I'm not sure what the right approach is:
* never create an explicit attribute for the TDR primary key. Regardless of the primary key column's name, Terra will autocreate a `{tablename}_id` column with this value. _(this is what's in this PR)_
* create an explicit attribute for the TDR primary key, unless that primary key is `{tablename}_id`, which Terra will autocreate

app/tests/test_tdr_manifest_to_rawls.py
jordansafer: Oops!

jordansafer:
> 



> 👍 on syntax, with one nit inline. I think this needs some PO input on behavior, though - I'm not sure what the right approach is:
> 
>     * never create an explicit attribute for the TDR primary key. Regardless of the primary key column's name, Terra will autocreate a `{tablename}_id` column with this value. _(this is what's in this PR)_
> 
>     * create an explicit attribute for the TDR primary key, unless that primary key is `{tablename}_id`, which Terra will autocreate

As a specific example:
Case 1: Table name is sample, primaryKey is aliquot_id. In Terra we add a sample_id column (with values taken from aliquot_id), and delete the aliquot_id column/attribute.

Case 2: Table name is sample, primaryKey is aliquot_id. In Terra we add a sample_id column (with values taken from aliquot_id), and keep the aliquot_id attribute too, so the values are shown twice.

*If the table name is sample, and the prmaryKey is sample_id, then we'll delete the column/attribute in both cases. Or we could prefix the column and use tdr:sample_id as an attribute with duplicate values.

app/translators/tdr_manifest_to_rawls.py
```diff
@@ -168,15 +173,22 @@ def create_attribute_value(self, value, is_reference: bool = False, reference_ta
...
-                and case_insensitive_name != case_insensitive_primary_key)
+                and case_insensitive_name[:-3] == table_name.lower())
+
+    @staticmethod
+    def attribute_name_is_primary_key(name: str, primary_key: str, table_name: str) -> bool:
+        case_insensitive_name = name.lower()
+        case_insensitive_primary_key = primary_key.lower() if primary_key is not None else None
+        return case_insensitive_name == case_insensitive_primary_key \
+            and case_insensitive_name.endswith('_id') \
+            and case_insensitive_name[:-3] == table_name.lower()
```
davidangb, line 192: the name of this method is now misleading, I think: it does far more than check if the attribute name is equal to the primary key name. Rename the method?

app/translators/tdr_manifest_to_rawls.py
```diff
@@ -125,6 +125,11 @@ def translate_parquet_row(self, row: pd.Series, column_names: List[str]) -> List
 
     def translate_parquet_attr(self, name: str, value) -> List[AttributeOperation]:
         """Convert a single cell of a pandas dataframe - assumed from a Parquet file - to an AddUpdateAttribute."""
+
+        # Don't add an attribute if this is the primary key and the name matches the table name
+        if ParquetTranslator.attribute_name_is_primary_key(name, self.table.primary_key, self.table.name): return []
+        if name == self.table.primary_key and name == self.table.name: return []
```
davidangb: the second line here - `if name == self.table.primary_key and name == self.table.name: return []` - is extraneous, isn't it? Can this line just be deleted? Also, the comment at line 129 is slightly incorrect.

davidangb (commented):
small comments inline, so close …

app/translators/tdr_manifest_to_rawls.py
jordansafer: Oops, forgot to delete! + rewording the comment to
\# Don't add an attribute if it's the primary key and it has the same name as {tableName}_id

app/translators/tdr_manifest_to_rawls.py
jordansafer: I see how it's confusing. Renaming to: attribute_should_be_skipped()

davidangb (approved):
thank you!!

sagehen03 (commented):
lgtm

sagehen03 (approved):

PR #98, opened 2022-04-06
AJ-354: Add api to mark imports as timed out after 36 hours in non-terminal state

sagehen03:
This just covers adding an end point that could be called on a schedule.  The new end point won't work until we add an enum entry for `TimedOut` which I've already done in dev.  

I didn't initially add auth for this end point since it doesn't reveal any data and import service is not open to the internet, but this is something we should agree on.


app/cleanup.py
```diff
@@ -0,0 +1,14 @@
...
+from app.db.model import Import, ImportStatus
+
+
+def clean_up_stale_imports():
+    with db.session_ctx() as sess:
+        stuck_jobs = sess.execute("""select id, status from imports where status NOT IN ('Error', 'Done', 'TimedOut') 
+        and HOUR(TIMEDIFF(NOW(), submit_time)) > 36""")
+        for job in stuck_jobs:
+            logging.warning(f"Job id {job['id']} has status {job['status']} but is more than 36 hours old")
+            Import.update_status_exclusively(job['id'], ImportStatus.from_string(job['status']), ImportStatus.TimedOut,
```
sagehen03: Few items for discussion here.. I could convert this to a single update statement rather than a select followed by updating one job at a time, but I think the total number of jobs will be small and I think it's nice to have the logging.  36 hours seems long?

app/db/model.py
```diff
@@ -56,6 +56,7 @@ class ImportStatus(enum.Enum):
     ReadyForUpsert = 50  # batchUpsert file has been put in bucket and rawls has been notified
     Upserting = 100  # rawls is actively working on importing the batchUpsert file
     Done = 500  # success
+    TimedOut = 99998  # https://broadworkbench.atlassian.net/browse/AJ-354
```
sagehen03: I couldn't discern if these status codes are used.  I didn't see anything in the database or in the code.  If I missed something then I might need to choose a non-arbitrary value.  If they're not used, should we switch to `auto()`?

app/cleanup.py
```diff
@@ -0,0 +1,15 @@
+from app.db import db
+import logging
+
+from app.db.model import Import, ImportStatus
+
+
+def clean_up_stale_imports():
+    with db.session_ctx() as sess:
+        # retrieve those jobs still in a 'transient/processing' state after more than 36 hours
+        stuck_jobs = sess.execute("""select id, status from imports where status NOT IN ('Error', 'Done', 'TimedOut')
```
jordansafer: I would push for defining a function in model.py to execute the query, something like get_stuck_jobs(), to keep the DB interactions in model.py. 

For the query, I think it looks great. I lean slightly towards SQL alchemy, with the tradeoff being that sqlalchemy would make it hard for us to take advantage of TIMEDIFF and other mysql functions, but that it would make it easier for us to migrate, if we changed underlying DBs to Postgres for example (which I think doesn't have TIMEDIFF). But I'm not emotional, I don't think there's any plan to migrate, and no performance concerns with the import service DB currently.

app/db/model.py
jordansafer: +1, the only place I see them used is in the DB. I'm worried about changing the number scheme + DB migration though, so I'd vote to stick with the current scheme-

app/server/routes.py
```diff
@@ -76,6 +76,14 @@ def get(self):
         """Return whether we and all dependent subsystems are healthy."""
         return health.handle_health_check(), 200
 
+@ns.route('/cleanup-jobs')
+class CleanUp(Resource):
+    @httpify_excs
+    @api.doc(security=None)
+    def get(self):
+        cleanup.clean_up_stale_imports()
+        return "ok"
```
jordansafer: nit, I think we should return 200 similar to line 77.

cron.yaml
```diff
@@ -0,0 +1,4 @@
+cron:
```
jordansafer, line 1: I love how clear and concise this is!

jordansafer (approved):

app/db/model.py
sagehen03: I don't think they are used in the database, the status column in the imports table is a string enum.

app/server/routes.py
sagehen03: I've added an explicit 200 response code which was returned implicitly before.

app/cleanup.py
sagehen03: I've moved the data access to model.py and switched to SQL Alchemy for the query mechanics.

sagehen03: Also I'm not a fan of ORMs, but we're using one already and I don't see our SQL needs being all that robust for this service anytime soon so the ORM potential for harm is low and I'd be in favor of moving to Postgres so consider me uneasily aligned with SQL Alchemy for now :)

https://blog.codinghorror.com/object-relational-mapping-is-the-vietnam-of-computer-science/

jordansafer (approved):
LGTM!

app/cleanup.py
jordansafer: I'm definitely a type 2 person in this list for most cases :)

app/db/model.py
```diff
@@ -51,12 +55,13 @@ def __init__(self, enum: Type[T]) -> None: ...
...
-    Upserting = 100  # rawls is actively working on importing the batchUpsert file
-    Done = 500  # success
-    Error = 99999  # something bad happened, check the error_message column for details
+    Pending = enum.auto()  # import request received by the user but we haven't done anything with it yet
+    Translating = enum.auto()  # in the process of translating to rawls batchUpsert
+    ReadyForUpsert = enum.auto()  # batchUpsert file has been put in bucket and rawls has been notified
+    Upserting = enum.auto()  # rawls is actively working on importing the batchUpsert file
+    Done = enum.auto()  # success
+    TimedOut = enum.auto()  # https://broadworkbench.atlassian.net/browse/AJ-354
+    Error = enum.auto()  # something bad happened, check the error_message column for details
```
davidangb, line 67: using `auto()` here is _probably_ ok, since these are listed in ascending order … but I think it's best to leave the enum values as explicit. We check the values here:
https://github.com/broadinstitute/import-service/blob/c8edba8a8a07128051d3dec30e25675ce94f6da4/app/status.py#L80
to ensure that status updates are incrementing.

app/db/model.py
```diff
@@ -152,13 +158,24 @@ def get(cls, import_id: str, sess: DBSession) -> Import:
         return sess.query(Import).filter(Import.id == import_id).one()
 
     @classmethod
-    def update_status_exclusively(cls, id: str, current_status: ImportStatus, new_status: ImportStatus, sess: DBSession) -> bool:
+    def get_stalled_imports(cls, sess: DBSession, job_age_hours: int) -> list[Import]:
+        """Retrieve those jobs still in a 'transient/processing' state after more than 36 hours."""
+        return sess.query(Import).filter(Import.status.notin_([ImportStatus.Error, ImportStatus.TimedOut,
+                                                               ImportStatus.Done]),
```
davidangb: can you use `ImportStatus.terminal_statuses()` here instead of listing out Error, TimedOut, Done?

app/cleanup.py
```diff
@@ -0,0 +1,12 @@
+from app.db import db, model
+import logging
+
+from app.db.model import Import, ImportStatus
+
+
+def clean_up_stale_imports(job_age_hours: int):
+    with db.session_ctx() as sess:
+        stuck_jobs = model.Import.get_stalled_imports(sess, job_age_hours)
```
davidangb, line 9: nit: if there exist > 0 stuck jobs, could you log the number of stuck jobs? You do have logging for each individual job, but it would be nice to see a single log line that shows the total number. We could even add alerts on this log entry - in theory we should always have 0 stuck jobs so we might want to know when we see any.

app/cleanup.py
```diff
@@ -0,0 +1,12 @@
...
+        for job in stuck_jobs:
+            logging.warning(f"Job id {job.id} has status {job.status} but is more than {job_age_hours} hours old")
```
davidangb: nit: could you log the job's `submit_time` here too, or even the timedelta between now and the job's `submit_time`?

davidangb (commented):
some comments inline!

I also feel a little wonky about leaving the `/cleanup-jobs` API unauthenticated … BUT since this is all still behind the VPN it feels safe. If/when we make Import Service public-facing we may have to reassess.

app/db/model.py
sagehen03: I added a comment that values need to be defined in processing order.  This set up is marginally more clear to me than increasing arbitrary numbers since we don't care about the values per se just that we transition from a smaller status to a larger one so I'd prefer to leave it in.

app/cleanup.py
davidangb: I had the same thoughts - agreed that a single update statement is technically more efficient, but the multiple statements give better visibility and we shouldn't have too many of them at once.

36 hours is definitely long, we could pull that back - that number was never intended to be a requirement, just a starting point

app/db/model.py
davidangb: I won't die on this hill, won't block the PR on this. I'm still a little nervous about the values changing underneath us:
https://docs.python.org/3/library/enum.html#using-automatic-values says the always-increment behavior of `auto()` "is an implementation detail and may change" …. though I doubt it will actually change in practice.

davidangb (approved):
one last optional comment on `auto()`, giving this a :+1: and your call on how to handle `auto()`

app/db/model.py
sagehen03: That's a fair point.  How about this for compromise https://broadworkbench.atlassian.net/browse/AJ-384 ?  Then those enum values won't matter anymore.

PR #100, opened 2022-04-28
AJ-392 Use Content-Range HEADER to get file size via GET before attempting download

sagehen03:


app/util/http.py
```diff
@@ -1,11 +1,31 @@
...
 from typing import IO, Iterator
 
+from app.constants import TWO_GB_IN_BYTES
+from app.util.exceptions import FileTooBigToDownlod
+
+
+def extractBytes(content_range):
+    # we expect value of the form `bytes 0-0/3268794`
+    # if for some reason we change the Range request header to pull additional bytes this pattern needs to change too
+    return int(re.sub(r'bytes 0-0/', '', content_range))
```
davidangb: 1. Could you make "0-0" into a variable and use it in both this `re.sub` and in the outgoing request header below? Then if for some reason we change the bytes we request it would only need to change in one place.
2. You could use the native string.replace here instead of a regex - `content_range.replace('bytes 0-0/', '')` … I doubt there's too much difference with such a simple replacement; your call

app/util/http.py
```diff
@@ -1,11 +1,31 @@
...
 import urllib.request
+import requests
 from contextlib import contextmanager
 from typing import IO, Iterator
 
+from app.constants import TWO_GB_IN_BYTES
+from app.util.exceptions import FileTooBigToDownlod
+
+
+def extractBytes(content_range):
```
davidangb: add the type hints?
```python
def extractBytes(content_range: str) -> int:
```

app/util/http.py
```diff
@@ -1,11 +1,31 @@
...
+    return int(re.sub(r'bytes 0-0/', '', content_range))
+
 
 @contextmanager
-def http_as_filelike(url: str) -> Iterator[IO]:
+def http_as_filelike(url: str, file_limit_bytes: int = TWO_GB_IN_BYTES) -> Iterator[IO]:
     """Open a file over HTTP and return it as a file-like object."""
+    headers = {"Range": "bytes=0-0"}
+    http_response = requests.get(url, headers=headers)
+    content_range = http_response.headers.get('Content-Range')
```
davidangb, line 23: based on our recent debugging experience, how about adding something like:
```python
    if http_response.status_code != 206:
        logging.error(f"Content-Range request returned {http_response.status_code} for {url} we won't download")
        raise FileTooBigToDownlod
```

app/util/http.py
```diff
@@ -1,11 +1,31 @@
...
+    if content_range is None:
+        logging.error(f"Content-Range header not returned by server for {url} we won't download")
+        raise FileTooBigToDownlod
+    if extractBytes(content_range) >= file_limit_bytes:
+        raise FileTooBigToDownlod
```
davidangb: 1. add a logging statement to this if clause too, to assist with any future debugging?
2. nit: typo in `FileTooBigToDownlod`, should be `FileTooBigToDownload`

davidangb (approved):
`extractBytes` could potentially have a unit test, but it's pretty small. :+1:

jordansafer (approved):

PR #101, opened 2022-05-03
PFB imports should create first-class Terra attribute lists [AJ-363]

davidangb:
AJ-363

If an inbound PFB contains property values that are arrays, translate those values into true attribute lists in Terra. Before this PR we translated those values into JSON representing the array.


app/tests/test_tdr_manifest_to_rawls.py
```diff
@@ -280,7 +280,7 @@ def find_add_update_attr(ops: Sequence[AttributeOperation], attrname: str) -> Ad
         return next(i for i in ops if isinstance(i, AddUpdateAttribute) and i.attributeName == attrname)
 
     def assert_attr_value(ops: Sequence[AttributeOperation], attrname: str, expected):
-        attr = find_add_update_attr(e.operations, attrname)
+        attr = find_add_update_attr(ops, attrname)
```
davidangb, line 283: not technically part of this PR. I was using this code as a basis for the new tests in this PR, and noticed that this line was incorrect, so I fixed it.

pytest.ini
```diff
@@ -1,3 +1,6 @@
 [pytest]
 filterwarnings =
-    ignore::DeprecationWarning:urllib3.*:
\ No newline at end of file
+    ignore::DeprecationWarning:urllib3.*:
+
+# Enable log display during test run (also known as “live logging”)
+log_cli = False
```
davidangb, line 6: `log_cli = False` is the default, so this doesn't change anything. But I found myself tweaking this value on and off and figured I'd leave it in here for future developers.

app/translators/pfb_to_rawls.py
```diff
@@ -43,14 +46,19 @@ def make_op(key, value):
...
+            else:
+                return [AddUpdateAttribute(key, value)]
 
         attributes = [make_op(key, value)
                       for key, value in record['object'].items() if value is not None]
         relations = [make_op(relation['dst_name'], EntityReference(entityType=relation['dst_name'], entityName=relation['dst_id']))
                      for relation in record['relations']]
 
-        return Entity(name, entity_type, [*attributes, *relations])
+        return Entity(name, entity_type, list(itertools.chain(*attributes, *relations)))
```
davidangb, line 61: this file is the only change to runtime code. The rest of this PR is additional testing.

pytest.ini
jordansafer: +1 :)

jordansafer (approved):
Looks great and clear cut :)

sagehen03 (approved):

PR #103, opened 2022-05-11
Aj 404 update cloud logging

calypsomatic:
AJ-404
- Updates google cloud logging from version 2.7 to 3.0.
- Removes workaround - no double logging seen. 
- Added custom json fields to logging in rawls to demonstrate possibility.  The old message was the entire http response.  Now it is just the message, with the rest of the response under 'jsonPayload'.

To generate json fields logging, try importing a snapshot to a workspace that doesn't exist.

I noticed two further changes from the update:

1. Relevant logs will be under the 'python' log.  Previously, they were found under 'app'.
logExplorer

2. Source method of the log used to be printed directly in the top level message

2 7nojson_fields

Now you must look inside for sourceLocation:

3 0withjson_fields_sourceLocation

This appears to be an automatic change, but you can display the source file and method by editing the summary display:

editsummary

### DIFF
--- a/app/external/rawls.py
+++ b/app/external/rawls.py
@@ -17,12 +17,13 @@ def get_workspace_uuid_and_project(workspace_namespace: str, workspace_name: str
         f"{os.environ.get('RAWLS_URL')}/api/workspaces/{workspace_namespace}/{workspace_name}?fields=workspace.workspaceId,workspace.googleProject",
         headers={"Authorization": bearer_token})
 
+    jso = resp.json()
     if resp.ok:
-        jso = resp.json()
         return RawlsWorkspaceResponse(workspace_id=jso["workspace"]["workspaceId"], google_project=jso["workspace"]["googleProject"])
     else:
         # just pass the error upwards
-        logging.info(f"Got {resp.status_code} from Rawls for {workspace_namespace}/{workspace_name}: {resp.text}")
+        logging.info(f"Got {resp.status_code} from Rawls for {workspace_namespace}/{workspace_name}: {jso['message']}",
+                     extra={"json_fields": jso})
         raise ISvcException(resp.text, resp.status_code)
 
 def check_workspace_iam_action(workspace_namespace: str, workspace_name: str, action: str, bearer_token: str) -> bool:
@@ -36,7 +37,9 @@ def check_workspace_iam_action(workspace_namespace: str, workspace_name: str, ac
         return False
     else:
         # just pass the error upwards
-        logging.info(f"Got {resp.status_code} from Rawls for {workspace_namespace}/{workspace_name}: {resp.text}")
+        jso = resp.json()
+        logging.info(f"Got {resp.status_code} from Rawls for {workspace_namespace}/{workspace_name}: {jso['message']}",
+                     extra={"json_fields": jso})
         raise ISvcException(resp.text, resp.status_code)
 
 
--- a/app/external/rawls.py
+++ b/app/external/rawls.py
@@ -22,7 +22,9 @@ def get_workspace_uuid_and_project(workspace_namespace: str, workspace_name: str
         return RawlsWorkspaceResponse(workspace_id=jso["workspace"]["workspaceId"], google_project=jso["workspace"]["googleProject"])
     else:
         # just pass the error upwards
-        logging.info(f"Got {resp.status_code} from Rawls for {workspace_namespace}/{workspace_name}: {resp.text}")
+        workspace_dict = { 'workspace': { 'namespace': workspace_namespace, 'name': workspace_name} }
+        logging.info(f"Got {resp.status_code} from Rawls for {workspace_namespace}/{workspace_name}: {resp.text}",
+                     extra={"json_fields": workspace_dict})
         raise ISvcException(resp.text, resp.status_code)
 
 def check_workspace_iam_action(workspace_namespace: str, workspace_name: str, action: str, bearer_token: str) -> bool:
--- a/app/external/rawls.py
+++ b/app/external/rawls.py
@@ -17,13 +17,12 @@ def get_workspace_uuid_and_project(workspace_namespace: str, workspace_name: str
         f"{os.environ.get('RAWLS_URL')}/api/workspaces/{workspace_namespace}/{workspace_name}?fields=workspace.workspaceId,workspace.googleProject",
         headers={"Authorization": bearer_token})
 
-    jso = resp.json()
     if resp.ok:
+        jso = resp.json()
         return RawlsWorkspaceResponse(workspace_id=jso["workspace"]["workspaceId"], google_project=jso["workspace"]["googleProject"])
     else:
         # just pass the error upwards
-        logging.info(f"Got {resp.status_code} from Rawls for {workspace_namespace}/{workspace_name}: {jso['message']}",
-                     extra={"json_fields": jso})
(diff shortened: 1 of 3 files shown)


davidangb:
> Relevant logs will be under the 'python' log. Previously, they were found under 'app'.

Is it still possible to navigate to the relevant logs by starting at the App Engine service (or version) in cloud console and clicking Tools -> Logs? I think it is, just want to confirm

calypsomatic:
Yep!

main.py
```diff
@@ -1,24 +1,9 @@
...
-    handler = AppEngineHandler(client, stream=sys.stderr)
-    handler.setFormatter(formatter)
-    handler.setLevel(logging.INFO)
-    root = logging.getLogger()
-    root.addHandler(handler)
-    root.setLevel(logging.INFO)
-else:
-    # For local runs, use normal Python logging (so we don't send all our test logs to Stackdriver!)
-    logging.basicConfig(format="%(module)s.%(funcName)s: %(message)s", level=logging.INFO)
-
```
sagehen03, line 21: I think we may still need this for FIABs :disappointed:

main.py
```diff
@@ -1,24 +1,9 @@
...
-    formatter = logging.Formatter("%(module)s.%(funcName)s: %(message)s")
-    handler = AppEngineHandler(client, stream=sys.stderr)
-    handler.setFormatter(formatter)
-    handler.setLevel(logging.INFO)
-    root = logging.getLogger()
-    root.addHandler(handler)
-    root.setLevel(logging.INFO)
-else:
-    # For local runs, use normal Python logging (so we don't send all our test logs to Stackdriver!)
-    logging.basicConfig(format="%(module)s.%(funcName)s: %(message)s", level=logging.INFO)
```
sagehen03, line 20: I think we may still need this for FIABs :disappointed: I get this error when I try to run the import service's docker file (which is used by FIABs) `Could not automatically determine credentials. Please set GOOGLE_APPLICATION_CREDENTIALS or explicitly create credentials and re-run the application. For more information, please see https://cloud.google.com/docs/authentication/getting-started` it could be that FIABs set the variable for us?  I'm not really sure.

calypsomatic:
How do I find out?

sagehen03 (commented):
Left a comment about potential FIAB complications.  Otherwise looks pretty good.  I see logging in cloud logging without dupes and I was able to do an import in dev

main.py
davidangb: Probably the quickest test here is just to run a manual swatomation test including this branch, and see what happens!

app/external/rawls.py
```diff
@@ -17,12 +17,13 @@ def get_workspace_uuid_and_project(workspace_namespace: str, workspace_name: str
         f"{os.environ.get('RAWLS_URL')}/api/workspaces/{workspace_namespace}/{workspace_name}?fields=workspace.workspaceId,workspace.googleProject",
         headers={"Authorization": bearer_token})
 
+    jso = resp.json()
```
davidangb: There are cases where the call to Rawls will not return json - a connection failure or a failure at the Rawls proxy layer, for instance. It looks like `resp.json()` will raise an exception if this happens, which would halt and not reach the logging statements below. Could you add some try/catch here, perhaps in the exception case you could manually craft a json object to be included in the logging?

main.py
calypsomatic: https://fc-jenkins.dsp-techops.broadinstitute.org/job/swatomation-pipeline/64494/
To me it looks like it built successfully and the failure was a test unrelated to my changes.  Either way, I guess I would vote for keeping that line in to keep fiab logging off the cloud.

app/external/rawls.py
```diff
@@ -36,7 +37,9 @@ def check_workspace_iam_action(workspace_namespace: str, workspace_name: str, ac
         return False
     else:
         # just pass the error upwards
-        logging.info(f"Got {resp.status_code} from Rawls for {workspace_namespace}/{workspace_name}: {resp.text}")
+        jso = resp.json()
+        logging.info(f"Got {resp.status_code} from Rawls for {workspace_namespace}/{workspace_name}: {jso['message']}",
+                     extra={"json_fields": jso})
```
davidangb: ah, same thing here, the response is not guaranteed to return json. Since this happens in multiple places, it looks like we could use a reusable function to safely extract json from a response object.

main.py
```diff
@@ -1,24 +1,9 @@
...
-
+client = google.cloud.logging.Client()
+client.setup_logging()
```
davidangb: given the previous code comment

> For local runs, use normal Python logging (so we don't send all our test logs to Stackdriver!)

Is this still an issue? for instance, do test runs or other local development send logs to Google as of this PR?

davidangb (commented):
some comments inline!

app/external/rawls.py
calypsomatic: So adding this json object to the logs was mainly to fulfill the task of demonstrating that it can be done - I looked at all the logging statements in the project and couldn't find a really good candidate, this was the best option.  Would it be better to just return it to its original state without the json_fields?

sagehen03 (approved):
Approved modulo David's comments about the test logging introducing a potential property does not exist error.  I think we could potentially just remove those changes.  We can validate that we're on version 3 without changing any existing log statements by seeing that the `sourceLocation` property is set in Google cloud logging (I see it already).

app/external/rawls.py
davidangb: I'm fine returning most everything to its original state, including moving the `resp.json()` back inside the `if resp.ok:` clause. See my suggestion below for proving structured logging support …

app/external/rawls.py
```diff
@@ -17,12 +17,13 @@ def get_workspace_uuid_and_project(workspace_namespace: str, workspace_name: str
...
 
+    jso = resp.json()
     if resp.ok:
-        jso = resp.json()
         return RawlsWorkspaceResponse(workspace_id=jso["workspace"]["workspaceId"], google_project=jso["workspace"]["googleProject"])
     else:
         # just pass the error upwards
-        logging.info(f"Got {resp.status_code} from Rawls for {workspace_namespace}/{workspace_name}: {resp.text}")
+        logging.info(f"Got {resp.status_code} from Rawls for {workspace_namespace}/{workspace_name}: {jso['message']}",
+                     extra={"json_fields": jso})
```
davidangb: since this is just a proof of structured logging support, how about something that doesn't risk exceptions, like:
```suggestion
        workspace_dict = { 'workspace': { 'namespace': workspace_namespace, 'name': workspace_name} }
        logging.info(f"Got {resp.status_code} from Rawls for {workspace_namespace}/{workspace_name}: {jso['message']}",
                     extra={"json_fields": workspace_dict})
```

main.py
davidangb: the error in that swatomation run DOES look unrelated to your changes - and it looks transient. BUT, it also prevented the Orchestration tests from running, and those are the tests that hit Import Service. So that run doesn't prove that this PR has no impact to FiaBs. Could you re-run?

calypsomatic:
https://fc-jenkins.dsp-techops.broadinstitute.org/job/swatomation-pipeline/64538/
This run seemed to work.

app/external/rawls.py
calypsomatic: Done.  And the swatomation run worked this time too (see below).

davidangb (approved):
lgtm, thank you!

PR #105, opened 2022-07-11
[AJ-456] Update to call new SAM API for sharing permissions in TDR import

jordansafer:
Update to call the latest SAM APIs to share permissions with TDR import. This allows us to share the snapshot with a workspace without requiring canShare permissions & without reading workspace emails from SAM.

This PR should not merge until after the new SAM API is merged and released as part of the monolith release.


davidangb (approved):
This all looks correct to me, I did not test it myself - I assume you did!

Looks like you have one codacy tweak :(

sagehen03 (approved):

PR #106, opened 2022-08-04
DR-2692 Allow reading from Azure HTTP endpoints

nmalfroy:
This PR enables imports from TDR exports for Azure snapshots

(Don't worry the PR isn't actually _that_ big...most of the line changes are test manifest files)

But basically, this allows reading from `gs://` _or `https://<>core.windows.net` and does the appropriate branching around reading from the gcs client or the http client.

It does some translation due to the nature of parquet files in TDR generated Azure parquet files (e.g. UUIDs are byte arrays and array fields are JSON string representations).

Permissions are handled via signed urls that get generated by TDR (both to point to the underlying parquet files and to point to the manifest itself)

I've manually tested this in the dev for both azure and gcp snapshots and they both seem to work

Also worth noting: I tried to keep the commits in meaningful chunks so reviewing my be easier that way


app/translators/tdr_manifest_to_rawls.py
```diff
@@ -113,14 +139,29 @@ def translate_data_frame(self, df: pd.DataFrame, column_names: List[str]) -> Ite
...
 
-        all_attr_ops = [self.translate_parquet_attr(colname, row[colname]) for colname in column_names]
+        all_attr_ops = []
+        for colname in column_names:
+            value = row[colname]
+            if is_azure is True:
+                # In Azure parquet files, the datarepo_row_id field is stored as bytes so we should convert to string
+                if colname == 'datarepo_row_id':
+                    value = str(uuid.UUID(bytes=value))
+                # In Azure parquet files, array fields are stored a stringified Json arrays that we should convert to arrays
```
jordansafer: nit: typo
```suggestion
                # In Azure parquet files, array fields are stored as stringified Json arrays that we should convert to arrays
```

app/translators/tdr_manifest_to_rawls.py
```diff
@@ -90,21 +109,28 @@ def translate(self) -> Iterator[Entity]:
...
     #     entity_batches = (self.translate_data_frame(b.to_pandas(), b.column_names) for b in reader)
     #     return itertools.chain(*entity_batches)
-    def translate_parquet_file_to_entities(self, file_like: IO) -> Iterator[Entity]:
+    def translate_parquet_file_to_entities(self, file_like: IO, is_azure: bool = False) -> Iterator[Entity]:
         """Converts single parquet file-like object to an iterator of Entity objects."""
-        pq_table: pyarrow.Table = pq.read_table(file_like)
+        pq_table: pyarrow.Table
+        if (is_azure):
+            # We need to read and wrap the bytes of the parquet file in a BytesIO object which implements methods required by the parquet reader
+            pq_table = pq.read_table(io.BytesIO(file_like.read()))
```
jordansafer, line 117: Instead of checking is_azure here, can we wrap the Azure file like in a BytesIO object as soon as we create it?

jordansafer (approved):

app/translators/tdr_manifest_to_rawls.py
nmalfroy: I was wondering about that.  I think the problem is that you end up having some duplicated logic since file_like gets created in two different areas (once in this class and once in the translate class (to handle the manifest))

davidangb: does this mean that the Parquet files are different for GCP exports and Azure exports, e.g. in how they represent datarepo_row_id?

Thinking out loud - I haven't tried to POC this - I wonder if we could have a base `ParquetTranslator` class, then specialize that to `GcpParquetTranslator` and `AzureParquetTranslator` and try to encapsulate the differences that way?

This comment shouldn't block the PR, I am going to thumb it

davidangb (approved):
👍 with the caveat that I haven't tried any of this myself in practice - I trust you will smoke-test in a real env!

app/translators/tdr_manifest_to_rawls.py
nmalfroy: Yeah, there are some subtle differences in how the parquet files look in the tow different cases (datarepo_row_id and arrays).   I guess I figured that this would be have the most code reuse since the two types are fairly similar

nmalfroy:
thanks!  I've been testing this on dev and it seems to do the thing.

PR #107, opened 2022-08-08
order tables for TDR import after ignoring invalid relationships [SUP-725]

davidangb:
For TDR imports, we order the imported tables so that Rawls will insert any _entities being referenced_ before it inserts the _entities doing the referencing_. We do this by inspecting the `relationships` key inside the TDR snapshot model.

However, we also only support a subset of the relationships that TDR supports: 
https://github.com/broadinstitute/import-service/blob/1e335fb32f470d6711ad11aae89869e6c1a9e4bc/app/external/tdr_manifest.py#L101

Before this PR, we considered all relationships when calculating the insertion order for tables. After this PR, we only consider those relationships which we support.

By considering fewer relationships, we hope to avoid instances of the "nodes are in a cycle" error from `TopologicalSorter`, which happen when the relationships are cyclical.


app/external/tdr_manifest.py
```diff
@@ -127,10 +129,11 @@ def get_ordering(relationships: List[Relationship]) -> List[str]:
         """
         ## Build a relationship graph of all the relationships
         relationship_graph = defaultdict(lambda : set())
-        for relationship in relationships:
-            from_table = relationship.from_.table
-            to_table = relationship.to.table
-            relationship_graph[from_table].add(to_table)
+        for table in tdr_tables:
+            for relationship in table.reference_attrs.values():
```
jordansafer: nit- let's call this a reference_attr, not a relationship. Then we'll have `to_table = reference_attr` on line 135

jordansafer (approved):

app/external/tdr_manifest.py
davidangb: good call, changed!

calypsomatic (approved):

PR #108, opened 2022-09-12
[AJ-537] Make TDR permissions sync optional and support triggering the permissions sync by param

jordansafer:
Note: We should not deploy this to production until the Terra-UI/orch changes have deployed.

This PR adds support for optionally syncing permissions in TDR snapshot import-by-copy. The importJob request will now contain an optional boolean field specifying whether the TDR permissions sync is enabled. The new flag is written to the import service (IS) DB as part of the job, and after the Rawls upsert completes, IS checks the flag to decide whether or not to sync permissions for TDR.

Tested end to end with TDR, Terra-UI, Orch, import service, rawls. I tested passing true, false, and no param paths, and verified that IS made API calls to sync permissions when the flag was enabled, and assumed false otherwise.


davidangb (approved):
lgtm! I did not test.

davidangb (approved):
:+1: again

sagehen03 (approved):

PR #112, opened 2022-09-23
Aj 568 cyclical tables

calypsomatic:
For [AJ-568](https://broadworkbench.atlassian.net/browse/AJ-568), allow cyclical tables in TDR snapshot-by-copy imports.  Previously, if a cycle was detected (i.e. table 1 references table 2 and table 2 references table 1), an error would be thrown.
Now, the import is marked as cyclical, and all reference attributes are reserved to the end of the import request, so that the tables have already been created before the references are added.
If an import is not cyclical, everything proceeds as before.


app/external/tdr_manifest.py
```diff
@@ -1,14 +1,10 @@
 from collections import defaultdict
 from dataclasses import dataclass
-from graphlib import TopologicalSorter
-from itertools import groupby
```
calypsomatic, line 4: These were just unused imports

app/tests/test_tdr_manifest_to_rawls.py
```diff
@@ -41,13 +43,13 @@ def test_translate_data_frame():
     assert len(entities) == 3
     assert entities[0].name == 'a'
     assert entities[0].entityType == 'unittest'
-    assert entities[0].operations == [AddUpdateAttribute('datarepo_row_id', 'a'), AddUpdateAttribute('one', 1), AddUpdateAttribute('two', 'foo'), _import_sourceid(random_snapshot_id), _import_timestamp(now)]
+    assert entities[0].operations == [_import_timestamp(now), _import_sourceid(random_snapshot_id), AddUpdateAttribute('datarepo_row_id', 'a'), AddUpdateAttribute('one', 1), AddUpdateAttribute('two', 'foo')]
```
calypsomatic, line 46: The change in processing attributes resulted in putting these bonus attributes (timestamp, sourceid) at the beginning instead of the end; it's slightly easier this way.

app/tests/test_tdr_manifest_to_rawls.py
```diff
@@ -334,3 +336,30 @@ def test_if_namespace_prefix_will_be_added():
...
+    jso = json.load(open('app/tests/resources/simple_cycle.json'))
+    monkeypatch.setattr(tdr_manifest_to_rawls.gcs, "open_file", open_fake_gcs_file)
+    parsed_manifest = TDRManifestParser(jso, fake_import_details.id)
+    tables = parsed_manifest.get_tables()
+    test = tmtr.translate_tables(fake_import_details, "snapshot_id", tables, parsed_manifest.is_cyclical())
+    import itertools
+    all_ops = [op for entity in list(itertools.chain(*test)) for op in entity.operations]
+    #The beginning of the list should be entirely non-references
+    first_ops = list(itertools.takewhile(lambda op:"_ref" not in op.attributeName, all_ops))
+    assert first_ops
```
davidangb, line 362: did you want some logic here to validate the contents of `first_ops`?

app/translators/tdr_manifest_to_rawls.py
```diff
@@ -47,16 +47,27 @@ def translate(self, import_details: Import, file_like: IO) -> Iterator[Entity]:
...
         pet_key = sam.admin_get_pet_key(import_details.workspace_google_project, import_details.submitter)
+        if not is_cyclical:
+            for t in tables:
+                for f in t.parquet_files:
+                    pt = ParquetTranslator(t, f, import_details, source_snapshot_id, pet_key, is_cyclical)
+                    yield pt.translate()
+        else:
+            yield from itertools.chain(TDRManifestToRawls.translate_ref_tables(import_details, source_snapshot_id, tables, True, pet_key, False), TDRManifestToRawls.translate_ref_tables(import_details, source_snapshot_id, tables, True, pet_key, True))
+    @classmethod
+    def translate_ref_tables(cls, import_details: Import, source_snapshot_id: str, tables: List[TDRTable],
```
davidangb: nit: the method name `translate_ref_tables` confused me at first; I initially assumed it would only translate the tables containing references; in actuality it translates all tables. Maybe something like `translate_table_parquet_files`?

app/translators/tdr_manifest_to_rawls.py
```diff
@@ -47,16 +47,27 @@ def translate(self, import_details: Import, file_like: IO) -> Iterator[Entity]:
...
+    def translate_tables(cls, import_details: Import, source_snapshot_id: str, tables: List[TDRTable], is_cyclical: bool) -> Iterator[Iterator[Entity]]:
         """Converts a list of TDR tables, each of which contain urls to parquet files, to an iterator of Entity objects."""
         pet_key = sam.admin_get_pet_key(import_details.workspace_google_project, import_details.submitter)
+        if not is_cyclical:
+            for t in tables:
+                for f in t.parquet_files:
+                    pt = ParquetTranslator(t, f, import_details, source_snapshot_id, pet_key, is_cyclical)
+                    yield pt.translate()
+        else:
+            yield from itertools.chain(TDRManifestToRawls.translate_ref_tables(import_details, source_snapshot_id, tables, True, pet_key, False), TDRManifestToRawls.translate_ref_tables(import_details, source_snapshot_id, tables, True, pet_key, True))
```
davidangb: I don't love that the `for t in tables:` loop is duplicated here and in `translate_ref_tables`. Could this method always delegate to `translate_ref_tables` and have that be the only place the loop is defined? I think the following would work, but I have not fully verified, please check my logic:

```suggestion
        if not is_cyclical:
            # when passing is_cyclical=False to translate_ref_tables, the value of translate_ref is irrelevant
            yield from itertools.chain(TDRManifestToRawls.translate_ref_tables(import_details, source_snapshot_id, tables, False, pet_key, False))
        else:
            yield from itertools.chain(TDRManifestToRawls.translate_ref_tables(import_details, source_snapshot_id, tables, True, pet_key, False), TDRManifestToRawls.translate_ref_tables(import_details, source_snapshot_id, tables, True, pet_key, True))
```

app/translators/tdr_manifest_to_rawls.py
```diff
@@ -160,9 +174,17 @@ def translate_parquet_row(self, row: pd.Series, column_names: List[str], is_azur
...
+            # For cyclical tables, we are either processing only the reference attributes or only the non-reference attributes at a time
+            if self.is_cyclical:
+                if ref_only and colname in self.table.reference_attrs:
+                    all_attr_ops.append(self.translate_parquet_attr(colname, value))
+                elif ref_only or colname in self.table.reference_attrs:
+                    continue
+                else:
+                    all_attr_ops.append(self.translate_parquet_attr(colname, value))
+            else:
+                all_attr_ops.append(self.translate_parquet_attr(colname, value))
```
davidangb: I think you can simplify this conditional:

```suggestion
            if self.is_cyclical and (colname in self.table.reference_attrs) != ref_only:
                continue
            else:
                all_attr_ops.append(self.translate_parquet_attr(colname, value))
```

app/tests/test_tdr_manifest_to_rawls.py
calypsomatic: I was thinking about it.  The main thing I'm looking for is that the first half is all non-ref and the second all ref.  Other unit tests validate that we get all the specific attributes we want so I really only need to check that once we find a ref value, everything else after it is also a ref value, which is covered in the rest of the test.  I could count the number since I know how many there would be rather than just checking it's not empty, but I don't know that it's necessary to look at the contents further than that they're not ref values.

davidangb: sounds good, that works for me. I was thrown in the PR between `assert first_ops` as a test to make sure `first_ops` is nonempty, vs. `assert first_ops` as a placeholder for forgotten code :)

davidangb (approved):

app/translators/tdr_manifest_to_rawls.py
```diff
@@ -47,16 +47,24 @@ def translate(self, import_details: Import, file_like: IO) -> Iterator[Entity]:
...
-        return itertools.chain(*self.translate_tables(import_details, source_snapshot_id, tables))
+        return itertools.chain(*self.translate_tables(import_details, source_snapshot_id, tables, parsed_manifest.is_cyclical()))
 
     @classmethod
-    def translate_tables(cls, import_details: Import, source_snapshot_id: str, tables: List[TDRTable]) -> Iterator[Iterator[Entity]]:
+    def translate_tables(cls, import_details: Import, source_snapshot_id: str, tables: List[TDRTable], is_cyclical: bool) -> Iterator[Iterator[Entity]]:
         """Converts a list of TDR tables, each of which contain urls to parquet files, to an iterator of Entity objects."""
         pet_key = sam.admin_get_pet_key(import_details.workspace_google_project, import_details.submitter)
+        if not is_cyclical:
+            yield from itertools.chain(TDRManifestToRawls.translate_table_parquet_files(import_details, source_snapshot_id, tables, False, pet_key, False))
```
jordansafer, line 57: nit: unless I'm missing something here- let's just pass `is_cyclical` as an argument, instead of wrapping these calls with an if-statement:
`yield from itertools.chain(TDRManifestToRawls.translate_table_parquet_files(import_details, source_snapshot_id, tables, is_cyclical, pet_key, False))`

jordansafer (approved):
one minor nit, otherwise looks good :)

app/translators/tdr_manifest_to_rawls.py
calypsomatic: If it's cyclical we need to make two separate calls and chain them together `itertools.chain(TDRManifestToRawls.translate_table_parquet_files(import_details, source_snapshot_id, tables, True, pet_key, False), TDRManifestToRawls.translate_table_parquet_files(import_details, source_snapshot_id, tables, True, pet_key, True))` instead of just one so I don't think that would work

PR #114, opened 2022-09-30
[AJ-608] fix attribute list name string

jordansafer:
https://broadworkbench.atlassian.net/browse/AJ-608

Imports for TDR snapshots with Attribute Lists are failing for references, the string appears to be incorrect


davidangb (approved):
:+1: looks correct in theory, I don't think it would make anything WORSE, will hopefully solve the current bug.

calypsomatic (approved):
👍 Assuming a manual test passes, but how could it be wrong?

PR #115, opened 2022-11-14
AJ-478: Include script and readme for deleting old import-service versions in GAE

aaronkanzer:
Related to https://broadworkbench.atlassian.net/browse/AJ-478


cleanup_scripts/delete-old-app-engine-version.sh
```diff
@@ -0,0 +1,95 @@
...
+if [ -z "${1+:}" ]; then
+    usage
+fi
+
+case $1 in
+    --help ) usage;;
+    dev|alpha|staging ) ;;
+    prod ) error "This script cannot be run against prod.";;
+    * ) error "ENV must be one of dev, alpha, or staging";;
+esac
```
davidangb, line 87: the usage for this script says it works against the perf environment, but this `case` check doesn't support perf. Could you add perf here?

Note: we are likely to delete the perf environment entirely, to stay in sync with broad-dsde-perf, but for now since terra-importservice-perf exists, I think this script should support it.

davidangb, line 87: additionally: this `case` check prevents the script from running against prod. What's your suggestion for how to clean up prod?

cleanup_scripts/delete-old-app-engine-versions-initial-cleanup.sh
```diff
@@ -0,0 +1,103 @@
...
+if [ -z "${1+:}" ]; then
+    usage
+fi
+
+case $1 in
+    --help ) usage;;
+    dev|alpha|staging ) ;;
+    prod ) error "This script cannot be run against prod.";;
+    * ) error "ENV must be one of dev, alpha, or staging";;
+esac
```
davidangb, line 89: same questions here as on the other script, regarding support for perf and prod

cleanup_scripts/delete-old-app-engine-version.sh
aaronkanzer: My suggestion would be that an authenticated user would go into the Google App Engine console and manage manually -- that way prod deletions are intentional, and if any bugs/changes to the `gcloud` CLI were introduced, it could not somehow affect `prod` -- let me know your thoughts.

I included a little blurb about this in the `README` section

aaronkanzer: Updated!

cleanup_scripts/delete-old-app-engine-versions-initial-cleanup.sh
aaronkanzer: Updated!

davidangb (approved):
:+1:. I haven't tested how well these scripts run, trusting you will do so!

jordansafer (approved):

PR #118, opened 2022-11-16
Re-arrange repo structure to include cleanup_scripts in Jenkins jobs

aaronkanzer:
Resolves error in https://fc-jenkins.dsp-techops.broadinstitute.org/blue/organizations/jenkins/general-deploy/detail/general-deploy/1221/pipeline


sagehen03 (approved):

davidangb (commented):
moving the `cleanup_scripts` directory into the `app` directory means that the scripts will get packaged up and deployed to App Engine, and the scripts will become part of the running app. I suppose that isn't a showstopper, but it's not ideal.

Some thoughts:
* is there anything in app.yaml or elsewhere during the deploy to GAE that we could use to exclude `cleanup_scripts` from being deployed?
* could you modify the `docker run` invocation to mount the `cleanup_scripts` into some other directory and run them from there, without holding them under `/app`?

davidangb:
ahh - take a look at the `.gcloudignore` file, I think that might help!

davidangb (approved):
once this successfully deploys to dev, would you double-check that the .gcloudignore is working as expected? I think you can do that by browsing the deployed source within Google Cloud Console.

aaronkanzer:
@davidangb will do!

PR #119, opened 2022-11-16
Revise bash header so Jenkins can execute script

aaronkanzer:


davidangb (approved):
might also try changing `deploy.sh` to have `./cleanup_scripts/delete-old-app-engine-version.sh` instead of `./cleanup_scripts/delete-old-app-engine-version` (note the explicit `.sh` in the first)

aaronkanzer:
@davidangb good call -- updated that as well

davidangb (approved):

calypsomatic (approved):

PR #126, opened 2023-03-15
AJ-876, AJ-900, AJ-570: Switch to poetry for dependency management, update all dependencies to relevant versions (including werkzeug==2.2.3), migrate from CircleCI to GHA, general cleanup

aaronkanzer:
Relates to https://broadworkbench.atlassian.net/browse/AJ-876, https://broadworkbench.atlassian.net/browse/AJ-900 and https://broadworkbench.atlassian.net/browse/AJ-570

- Introduces `poetry` as a dependency manager into `import-service` -- https://python-poetry.org/ (`Poetry` is great in terms of resolving complex dependencies between multiple Python libraries at once -- this is important for `import-service`, as a simple `requirements.txt`/using `pip` does not do similar logic under the hood.)

- migrates `import-service` from `CircleCI` to `Github Actions` -- `poetry` is better supported (and frankly operational) in Github Actions.

- Updates SQL calls in couple places with `text(` prefix for SQLAlchemy > 2 updates

- Removes references to `git-secrets` thus removing minni-kenny scripts


.github/workflows/publish-docker.yml
```diff
@@ -0,0 +1,64 @@
...
+
+      - name: Construct GCR docker image name
+        id: gcr-image-name
+        run: echo "name=${GCR_REGISTRY}/${GOOGLE_PROJECT}/${SERVICE_NAME}" >> $GITHUB_OUTPUT
+
+      - name: Set up Docker Buildx
+        uses: docker/setup-buildx-action@v1
+
+      - name: Build Docker image
+        run: docker build . -t ${{ steps.gcr-image-name.outputs.name }}:${{ steps.setHash.outputs.git_short_sha }}
```
davidangb: the previous Circle build used a tag of `${CIRCLE_TAG:-${CIRCLE_BRANCH}}`, this uses a tag of `$(git rev-parse --short HEAD)` … are there any important differences? I don't care if we switch to a new tagging scheme, just wanna make sure we're aware of what we're changing.

Similarly, I think the Circle build tagged `latest` but the GHA build doesn't?

.github/workflows/publish-docker.yml
```diff
@@ -0,0 +1,64 @@
...
+# TODO: refine if we need this
+#on:
+#  workflow_call:
+#    inputs:
+#      new-tag:
+#        required: true
+#        type: string
+#    secrets:
+#      BROADBOT_TOKEN:
+#        required: true
```
davidangb: probably don't need this, unless we want to call this from another workflow

.github/workflows/publish-docker.yml
aaronkanzer: Ah I should've prefaced this draft with me using a unique value (wanted to make sure I didn't overwrite anything), but yes, we should eventually use `latest` :D

.github/workflows/publish-docker.yml
aaronkanzer: Removed!

app/tests/test_tdr_manifest_to_rawls.py
```diff
@@ -340,7 +340,7 @@ def test_if_namespace_prefix_will_be_added():
 
 def get_fake_cyclic_parquet_translator(table: TDRTable) -> ParquetTranslator:
     fake_import_details = Import('workspace_name:', 'workspace_ns', 'workspace_uuid', 'workspace_google_project', 'submitter', 'import_url', 'filetype', True)
-    return ParquetTranslator(table, "doesntmatter", fake_import_details, 'source_snapshot_uuid', None, True)
+    return ParquetTranslator(table, "doesntmatter", fake_import_details, 'source_snapshot_uuid', None, True)  # type: ignore
```
aaronkanzer, line 343: `mypy` > 1, and `sqlalchemy` > 2 don't play nicely 100% together

The `# type: ignore` inline comments are to resolve a warning of something in nature to: `Incompatible default for argument "auth_key" (default has type "None", argument has type "Dict[str, Any]")`

This _isn't_ actually problematic, just `mypy` has a problem with it

Dockerfile
```diff
@@ -1,18 +1,31 @@
...
-# Install Python dependencies into the virtualenv
-COPY requirements.txt /
-RUN /venv/bin/pip install -r /requirements.txt
-RUN /venv/bin/pip install gunicorn
+# Install poetry separated from system interpreter
+RUN python3 -m venv $POETRY_VENV \
+    && $POETRY_VENV/bin/pip install -U pip setuptools \
+    && $POETRY_VENV/bin/pip install poetry==${POETRY_VERSION}
 
-FROM us.gcr.io/broad-dsp-gcr-public/base/python:3.9-debian
```
aaronkanzer, line 10: It seems that both `-alpine` and `-debian` fail our builds, but standard `python:3.9` does not -- going to work with AppSec to see how I can resolve this

aaronkanzer: Spoke with @davidangb -- this `Dockefile` isn't used in prod, thus we don't need to use an app-sec blessed image necessarily.

mypy.sh
jordansafer, line 1: thoughts on keeping this vs the longer command you have to run/remember without it? it's definitely convenient to have a script instead of going back to the README each time to get the syntax

jordansafer (approved):

poetry.lock
```diff
@@ -0,0 +1,2702 @@
...
+]
+
+[package.extras]
+brotli = ["brotli (>=1.0.9)", "brotlicffi (>=0.8.0)", "brotlipy (>=0.6.0)"]
+secure = ["certifi", "cryptography (>=1.3.4)", "idna (>=2.0.0)", "ipaddress", "pyOpenSSL (>=0.14)", "urllib3-secure-extra"]
+socks = ["PySocks (>=1.5.6,!=1.5.7,<2.0)"]
+
+[[package]]
+name = "werkzeug"
+version = "2.2.3"
```
aaronkanzer, line 2564: Confirmation that we are using `werkzeug==2.2.3`

davidangb (commented):
A few questions … 
1. Should this be tagging docker builds with `:latest`?
2. Why was git-secrets/minnie-kenny removed?
3. Is the docker build action working correctly? I see in the latest run it's been going for more than 2 hours …

aaronkanzer:
> A few questions …
> 
> 1. Should this be tagging docker builds with `:latest`?
> 2. Why was git-secrets/minnie-kenny removed?
> 3. Is the docker build action working correctly? I see in the latest run it's been going for more than 2 hours …

- Updated back to `latest` -- was just using something unique to not overwrite things for testing
- Put `git-secrets` and `minnie-kenny` back
- Docker builds should be fixed now!

Dockerfile
```diff
@@ -1,18 +1,29 @@
...
+
+RUN poetry lock
+RUN poetry install --no-cache
+RUN poetry add gunicorn
+
 COPY . .
 
 EXPOSE 8080
-
-CMD /venv/bin/gunicorn -b :8080 main:app
```
aaronkanzer, line 18: This is now invoked in the `terra-helmchart` directly for `import-service`

.github/workflows/publish-docker.yml
```diff
@@ -0,0 +1,44 @@
...
+        run: echo "name=${GCR_REGISTRY}/${GOOGLE_PROJECT}/${SERVICE_NAME}" >> $GITHUB_OUTPUT
+
+      - name: Set up Docker Buildx
+        uses: docker/setup-buildx-action@v1
+
+      - name: Build Docker image
+        run: docker build . -t ${{ steps.gcr-image-name.outputs.name }}:latest
+
+      - name: Push Docker image
+        run: docker push ${{ steps.gcr-image-name.outputs.name }}:latest
```
davidangb: is this _only_ tagging with `latest` now? no git hashes? Take a look at the Circle workflow - it would tag every docker build with a git hash/branch, and then would ALSO tag with `latest` (if it was running on the `develop`) branch. I don't think we want to lose the git hash/branch names.

.github/workflows/publish-docker.yml
```diff
@@ -0,0 +1,44 @@
+name: Publish Docker image to GCR
+
+on:
+  push:
+    branches:
+      - develop
```
davidangb: I'm guessing we want to build images for branches/PRs as well. Otherwise, how would one test a branch in a BEE?

Dockerfile
davidangb, line 18: Even though the `terra-helmfile` PR overrides this, any reason to take it out? Leaving it in (and updating it to use `poetry`) could allow some flexibility if anyone wanted to run the container standalone. I don't feel strongly about this either way; feel free to ignore this comment.

.github/workflows/publish-docker.yml
aaronkanzer: Would putting some documentation -- e.g. a `Running Locally or within a BEE` section be sufficient?

My concern is that we would push a handful of images to GCR that would be unnecessary to do? Let me know your thoughts.

.github/workflows/publish-docker.yml
aaronkanzer: Updated this step to have an `if` else` clause that matches what we used prior in CircleCI in terms of tagging name convention

Dockerfile
aaronkanzer: I put the `CMD` value back in

requirements.txt
```diff
@@ -1,26 +1,113 @@
...
-google-cloud-pubsub==2.9.0
-pypfb==0.5.21
-gcsfs==2021.11.1
-memunit==0.5.2
-pyhumps==3.0.2
-flask-restx==0.5.1
-pyarrow==6.0.1
-pydantic==1.8.2
-pandas==1.5.3
+aiofiles==0.8.0 ; python_version >= "3.9" and python_version < "3.10"
```
aaronkanzer, line 2: This file is included specifically for Google App Engine requirements

.github/workflows/publish-docker.yml
davidangb: This still looks like it changes behavior compared to the Circle builds. Is that intentional? What I see is:

Circle
* always build the docker image with a hash/branch tag ("Build image" step)
* if on `develop`, add an additional tag of `latest`("Build image" step)
* always push to the hash/branch tag ("Push to GCR" step)
* if on `develop`, also push to the `latest` tag ("Push to GCR" step)

GHA
* if on `develop`, build and push to `latest` ("Build and push Docker image" step)
* if not on `develop`, build and push to the hash ("Build and push Docker image" step)
* always push to the `latest` tag ("Push Docker image" step - I think this was left in by mistake)

The main difference is that in GHA, builds on the `develop` branch no longer get tagged with a hash. I think this would be a loss, as it prevents us from looking at the list of commits and grabbing an image that matches a previous commit. If you're changing this intentionally, I'm open to suggestions on why it's better!

davidangb: also, given the `on: … push: … branches: … - develop` trigger at the top of the file, the `else` clause here will never fire.

.github/workflows/publish-docker.yml
davidangb: What are you thinking for a process here? If a user wants to run a PR branch locally or in a bee, would they have to manually build the image? If so, how about adding `workflow_dispatch` as an action trigger here, so a developer could manually kick off this action instead of building on their own machine?

requirements.txt
davidangb: maybe add a code comment that this file is generated by poetry (I assume this is the case) and that developers should not edit it manually

.github/workflows/publish-docker.yml
aaronkanzer: I was seeing the image pushed (if within the `develop` branch) in CircleCI tagged as `us.gcr.io/broad-dsp-gcr-public/import-service:lastest`

See here: https://github.com/broadinstitute/import-service/blob/develop/.circleci/config.yml#L121

Perhaps this is where I am slightly confused. I agree that using the hash value is better; however, this seems it would deviate from the pattern in CircleCI?

.github/workflows/publish-docker.yml
aaronkanzer: Just tying this in with https://github.com/broadinstitute/import-service/pull/126#discussion_r1146767243

Good point on the `workflow_dispatch` -- I added that as a requirement for the file in general, just to make it intentional, and to conserve the amount of pushes we do to GCR.

.github/workflows/publish-docker.yml
aaronkanzer: `always push to the latest tag ("Push Docker image" step - I think this was left in by mistake)` yes this was left in by mistake ha -- removed!

requirements.txt
aaronkanzer: docstring added at top!

.github/workflows/publish-docker.yml
davidangb: > I was seeing the image pushed (if within the develop branch) in CircleCI tagged as us.gcr.io/broad-dsp-gcr-public/import-service:lastest

Apologies for the (long) runaround here. I'm looking at the actual images in `us.gcr.io` and they're not what I expected from reading the code. All the "extra" images e.g. `us.gcr.io/broad-dsp-gcr-public/import-service@sha256:3047ac2b0687fc32287818dd8c0a279f31228deaa3d698b0bac4452b93c15186` don't have the hash tag I expected. You're right about that.

The only difference I see in practice is that Circle will tag with `develop` and `latest`, whereas this GHA code will only tag with `latest` - it would be nice to replicate that but I don't care too much.

davidangb (approved):
ty and sorry for my misunderstanding on tags!

PR #127, opened 2023-04-04
AJ-942: Include support for anvil.gi.ucsc.edu URL in import service

aaronkanzer:
It was requested that `https://*.anvil.gi.ucsc.edu/manifest/files` be an acceptable URL for files imported via `import-service`

This PR introduces that logic, as well as a mini unit-test to ensure the functionality will work correctly the first time

### DIFF
--- a/app/tests/test_path_validation.py
+++ b/app/tests/test_path_validation.py
@@ -2,11 +2,11 @@ import flask.testing
 import pytest
 import os
 
-from app.tests import testutils
-from app import translate
-from app.auth import userinfo
-from app.db.model import *
+from app.auth.userinfo import UserInfo
+from app.translate import validate_import_url
 
+from app import translate
+from app.util.exceptions import InvalidPathException
 
 good_headers = {"Authorization": "Bearer ya29.blahblah"}
 
@@ -18,11 +18,14 @@ def assert_response_code_and_logs(resp, caplog, import_url):
     auditlog = filter(lambda rec: rec.message == f"User 123456 hello@bees.com attempted to import from path {import_url}", caplog.records)
     assert list(auditlog), "Expected audit log message to exist if user specified illegal domain; did not find such message in log."
 
+user_info = UserInfo("subject-id", "awesomepossum@broadinstitute.org", True)
 @pytest.mark.parametrize("netloc", translate.VALID_NETLOCS)
 @pytest.mark.parametrize("filetype", translate.FILETYPE_TRANSLATORS.keys())
 @pytest.mark.usefixtures("sam_valid_user", "user_has_ws_access", "pubsub_publish", "pubsub_fake_env")
 def test_legal_netlocs_simple(client, netloc, filetype):
-    payload = {"path": f"https://{netloc}/some/valid/path", "filetype": filetype}
+    path = f"https://{netloc}/some/valid/path"
+    payload = {"path": path, "filetype": filetype}
+    assert validate_import_url(path, filetype, user_info)
     resp = client.post('/namespace/name/imports', json=payload, headers=good_headers)
     assert resp.status_code == 201
     # NB we don't test anything deeper than the 201 response; other tests check to see if the
@@ -31,7 +34,10 @@ def test_legal_netlocs_simple(client, netloc, filetype):
 @pytest.mark.usefixtures("sam_valid_user", "user_has_ws_access", "pubsub_publish", "pubsub_fake_env")
 @pytest.mark.parametrize("filetype", translate.FILETYPE_TRANSLATORS.keys())
 def test_illegal_netloc_simple(client: flask.testing.FlaskClient, filetype, caplog):
-    payload = {"path": "https://haxxor.evil.bad/some/valid/path", "filetype": filetype}
+    path = "https://haxxor.evil.bad/some/valid/path"
+    payload = {"path": path, "filetype": filetype}
+    with pytest.raises(InvalidPathException):
+        validate_import_url(path, filetype, user_info)
     resp = client.post('/namespace/name/imports', json=payload, headers=good_headers)
     assert_response_code_and_logs(resp, caplog, payload["path"])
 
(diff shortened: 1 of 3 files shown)


app/tests/test_translate.py
```diff
@@ -349,3 +352,19 @@ def test_bad_gcs(fake_import, fake_publish_rawls, client):
 
     # no pubsub message should have been sent
     fake_publish_rawls.assert_not_called()
+
+
+user_info = UserInfo("subject-id", "awesomepossum@broadinstitute.org", True)
+@pytest.mark.parametrize("import_url, is_valid", [
```
aaronkanzer, line 358: While this doesn't build out tests for all logic in `validate_import_url`, just wanted to add some tests to ensure these changes work, without increasing scope since this functionality is helpful immediately to UCSC

calypsomatic (approved):

app/tests/test_translate.py
```diff
@@ -349,3 +352,19 @@ def test_bad_gcs(fake_import, fake_publish_rawls, client):
...
+    ("https://something.anvil.gi.ucsc.edu", True),
+    ("something.anvil.gi.ucsc.edu", False),
+])
+def test_validate_import_url(import_url, is_valid):
+    for file_type in FILETYPE_TRANSLATORS:
+        if not is_valid:
+            with pytest.raises(InvalidPathException):
+                validate_import_url(import_url=import_url, import_filetype=file_type, user_info=user_info)
+        else:
+            assert validate_import_url(import_url=import_url, import_filetype=file_type, user_info=user_info) is is_valid
```
davidangb, line 371: you might be able to combine this into `test_legal_netlocs_simple` and/or `test_illegal_netloc_simple`:
https://github.com/broadinstitute/import-service/blob/618fe740d778d70782be6fb2ca784fda56725eb6/app/tests/test_path_validation.py#L21-L36

davidangb (commented):
one comment about consolidating unit tests, the runtime change looks good!

app/tests/test_translate.py
aaronkanzer: Added a couple more asserts in those unit tests!

app/tests/test_path_validation.py
```diff
@@ -18,11 +18,14 @@ def assert_response_code_and_logs(resp, caplog, import_url):
     auditlog = filter(lambda rec: rec.message == f"User 123456 hello@bees.com attempted to import from path {import_url}", caplog.records)
     assert list(auditlog), "Expected audit log message to exist if user specified illegal domain; did not find such message in log."
 
+user_info = UserInfo("subject-id", "awesomepossum@broadinstitute.org", True)
 @pytest.mark.parametrize("netloc", translate.VALID_NETLOCS)
```
davidangb: could you do something like:
```python
@pytest.mark.parametrize("netloc", translate.VALID_NETLOCS +
                         ["something.anvil.gi.ucsc.edu/manifest/files",
                            "something-else.anvil.gi.ucsc.edu/manifest/files",
                            "*.anvil.gi.ucsc.edu/manifest/files",
                            "something.anvil.gi.ucsc.edu"])
```
here, and eliminate the `test_validate_import_url` in test_translate.py entirely? I'd prefer that over having the `validate_import_url` test cases spread across two files.

You'd want to do something similar for `test_illegal_netloc_simple` if you want to add negative test cases.

aaronkanzer: updated -- I think I'm OK with the positive cases here for now -- thanks

app/tests/test_path_validation.py
```diff
@@ -18,11 +18,17 @@ def assert_response_code_and_logs(resp, caplog, import_url):
     auditlog = filter(lambda rec: rec.message == f"User 123456 hello@bees.com attempted to import from path {import_url}", caplog.records)
     assert list(auditlog), "Expected audit log message to exist if user specified illegal domain; did not find such message in log."
 
-@pytest.mark.parametrize("netloc", translate.VALID_NETLOCS)
+
+user_info = UserInfo("subject-id", "awesomepossum@broadinstitute.org", True)
```
davidangb, line 22: nit: I think `user_info` is unused (and when deleting this also delete the import, assuming that's unused too)

app/tests/test_path_validation.py
```diff
@@ -2,11 +2,11 @@
...
 
-from app.tests import testutils
-from app import translate
-from app.auth import userinfo
-from app.db.model import *
+from app.auth.userinfo import UserInfo
+from app.translate import validate_import_url
 
+from app import translate
+from app.util.exceptions import InvalidPathException
```
davidangb, line 9: nit: unused import. I thought `mypy` would catch this!

app/tests/test_translate.py
```diff
@@ -349,3 +352,20 @@ def test_bad_gcs(fake_import, fake_publish_rawls, client):
...
+user_info = UserInfo("subject-id", "awesomepossum@broadinstitute.org", True)
+@pytest.mark.parametrize("import_url, is_valid", [
+    ("https://something.anvil.gi.ucsc.edu/manifest/files", True),
+    ("https://something-else.anvil.gi.ucsc.edu/manifest/files", True),
+    ("https://*.anvil.gi.ucsc.edu/manifest/files", True),
+    ("https://something.anvil.gi.ucsc.edu", True),
+    ("something.anvil.gi.ucsc.edu", False),
+])
+def test_validate_import_url(import_url, is_valid):
+    for file_type in FILETYPE_TRANSLATORS:
```
davidangb, line 366: I always prefer to do this kind of loop via `@pytest.mark.parametrize` instead of looping inside the body of the test.

When there's a loop inside the body of the test, the test will stop executing at the first failure and skip the remainder of the loop. If you use `@pytest.mark.parametrize`, it will continue executing all permutations and report all failures.

I won't hold up the PR for this; it's not a major issue, but figured I'd bring it up for the sake of discussion.

davidangb (approved):
A few comments inline about unused vars/imports, otherwise looks good, and I'll give this a proactive :+1:. If all you change is removing the unused stuff, I don't need to re-review. Thanks!

PR #130, opened 2023-06-07
AJ-1105 Identify protected data

calypsomatic:
[AJ-1105](https://broadworkbench.atlassian.net/browse/AJ-1105)
If an import is from a protected system, identify it as such.  At this time we are only considering pfb imports from AnVIL as protected data.  This PR inspects the source URL and filetype to see that it matches our definition of protected, but acts no further; behavior changes based on the protected status of the data will come in a later PR.
I've placed the protected data check in the `validate_import_url()` method in `translate.py` in order to reuse parsing and validation of the URL.  I waffled between that double-parsing and -validating. Another option would be to double-parse but after validation, from the `handle()` method in `new_import.py`.  This choice depends in part on where and how we plan to react to the protected status of the data later on, but it's relatively easy to refactor as needed.

[AJ-1105]: https://broadworkbench.atlassian.net/browse/AJ-1105?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


app/translate.py
```diff
@@ -31,6 +31,8 @@
 
 VALID_TDR_SCHEMES = ["gs", "https"]
 
+PROTECTED_NETLOCS = ["anvil.gi.ucsc.edu", "anvilproject.org"]
+
```
davidangb, line 36: adding "anvilproject.org" here is misleading - an import from "anvilproject.org" will be rejected because that domain isn't in `VALID_NETLOCS`. How about something like:
```python
PROTECTED_NETLOCS = ["anvil.gi.ucsc.edu", "anvilproject.org"]
VALID_NETLOCS = PROTECTED_NETLOCS + ["s3.amazonaws.com", "storage.googleapis.com", "service.azul.data.humancellatlas.org", "dev.singlecell.gi.ucsc.edu", "core.windows.net"]
```

app/translate.py
```diff
@@ -173,3 +177,10 @@ def validate_import_url(import_url: Optional[str], import_filetype: Optional[str
     else:
         logging.warning(f"Unrecognized netloc or bucket for import: [{parsedurl.netloc}] from [{import_url}]")
         raise exceptions.InvalidPathException(import_url, user_info, "File cannot be imported from this URL.")
+
+def is_protected(import_netloc: str, import_filetype: Optional[str]) -> bool:
+    """Determines whether an import is protected data based on where it's imported from
+    and its filetype.  Initially, only PFBs from AnVIl are considered protected data"""
```
davidangb, line 184: thank you for the attention to it's and its :)

davidangb (commented):
one comment on netlocs, looking good! I think the location for `is_protected` is just fine.

davidangb (approved):
looks good!

davidangb (approved):
still :+1:

yuliadub (approved):

PR #132, opened 2023-07-05
AJ-1104 do not import protected data into unprotected workspace

calypsomatic:
[AJ-1104](https://broadworkbench.atlassian.net/browse/AJ-1104)
If the data being imported is protected (as determined by its host, implemented in https://github.com/broadinstitute/import-service/pull/130), but the workspace is not (no auth domain and bucket not marked as "secure"), then throw an error instead of importing.
To avoid parsing the URL twice, I've modified the `validate_import_url` method to return the url's host instead of a boolean; the boolean was not actually used; if the method did not return True then it simply threw an error.  I also moved this method from `translate.py` to `new_import.py` to make things slightly easier and because that seemed more appropriate to me.  This required a few other changes throughout the code.
The `RawlsWorkspaceResponse` has also been modified to contain the auth domains and bucket for validation, and I've updated the mock `RawlsWorkspaceResponse` to have a secure bucket to prevent unrelated tests from failing at this validation step.

[AJ-1104]: https://broadworkbench.atlassian.net/browse/AJ-1104?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


davidangb (approved):
looks good to me, thank you!

app/tests/test_status.py
```diff
@@ -1,13 +1,13 @@
 import pytest
 import unittest.mock as mock
 
-from app import translate
+from app import translate, new_import
```
yuliadub: doesnt look like translate is used anymore

yuliadub (approved):

app/tests/test_status.py
calypsomatic: good catch!

PR #138, opened 2023-08-03
[AJ-1129] Validate GCS URLs in tdrexport requests

nawatts:
https://broadworkbench.atlassian.net/browse/AJ-1129

Currently, when a `tdrexport` import is created with a GCS URL, the only validation done on the URL is that it starts with `gs://`.

https://github.com/broadinstitute/import-service/blob/63c1c98a5851c064a4150a820e476f164189502a/app/new_import.py#L117-L118

Note that Python's `urlparse` does _not_ perform validation.
https://docs.python.org/3.9/library/urllib.parse.html#url-parsing-security

This adds validation of the URL using Pydantic's [AnyUrl](https://docs.pydantic.dev/1.10/usage/types/#urls) type.

### DIFF
--- a/app/new_import.py
+++ b/app/new_import.py
@@ -82,9 +82,12 @@ def is_protected_workspace(authorization_domain: Optional[Set[str]], bucket_name
     return bucket_name and bucket_name.startswith("fc-secure")
 
 @validate_arguments
-def validate_and_parse_url(url: AnyUrl) -> ParseResult:
+def _validate_url(url: AnyUrl) -> str:
+    return str(url)
+
+def validate_and_parse_url(url: str) -> ParseResult:
     """Validates the input URL using pydantic and parses it with urlparse."""
-    return urlparse(str(url))
+    return urlparse(_validate_url(url)) # type: ignore
 
 def validate_import_url(import_url: Optional[str], import_filetype: Optional[str], user_info: UserInfo) -> str:
     """Inspects the URI from which the user wants to import data. Because our service will make an
--- a/app/new_import.py
+++ b/app/new_import.py
@@ -7,8 +7,9 @@ from app.external import sam, pubsub
 from app.auth import user_auth
 from app.util import exceptions
 
+from pydantic import AnyUrl, validate_arguments
 from typing import Optional, Set
-from urllib.parse import urlparse
+from urllib.parse import ParseResult, urlparse
 import os
 
 from app.auth.userinfo import UserInfo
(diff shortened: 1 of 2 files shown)


app/tests/test_path_validation.py
```diff
@@ -20,7 +20,6 @@ def assert_response_code_and_logs(resp, caplog, import_url):
 user_info = UserInfo("subject-id", "awesomepossum@broadinstitute.org", True)
 @pytest.mark.parametrize("netloc", new_import.VALID_NETLOCS + ["something.anvil.gi.ucsc.edu/manifest/files",
                                                               "something-else.anvil.gi.ucsc.edu/manifest/files",
-                                                              "*.anvil.gi.ucsc.edu/manifest/files",
```
nawatts, line 23: This test was making requests with invalid URLs. This was added in #127.

app/new_import.py
```diff
@@ -80,6 +81,14 @@ def is_protected_workspace(authorization_domain: Optional[Set[str]], bucket_name
         return True
     return bucket_name and bucket_name.startswith("fc-secure")
 
+@validate_arguments
+def _validate_url(url: AnyUrl) -> str:
+    return str(url)
+
+def validate_and_parse_url(url: str) -> ParseResult:
+    """Validates the input URL using pydantic and parses it with urlparse."""
+    return urlparse(_validate_url(url)) # type: ignore
```
nawatts, line 92: Mypy expects `_validate_url` to take an `AnyUrl` argument, but it actually can accept a string that pydantic coerces into an `AnyUrl`. Breaking this up into two functions keeps the `type: ignore` confined here.

app/tests/test_path_validation.py
```diff
@@ -46,6 +45,12 @@ def test_unparsable_path(client: flask.testing.FlaskClient, caplog):
     resp = client.post('/namespace/name/imports', json=payload, headers=good_headers)
     assert_response_code_and_logs(resp, caplog, payload["path"])
 
+@pytest.mark.usefixtures("sam_valid_user", "user_has_ws_access", "pubsub_publish", "pubsub_fake_env")
+def test_invalid_tdrexport_gcs_url(client):
+    payload = {"path": "gs://bucket https://example.com/foo", "filetype": "tdrexport"}
```
calypsomatic: Do we need to test any other formats as invalid?

nawatts: Other `filetypes` / `path` schemes were getting _some_ additional validation out of being compared to `VALID_NETLOCS`, but yeah, they should be checked for valid URLs as well.

nawatts: Added a few more tests in https://github.com/broadinstitute/import-service/pull/138/commits/e7ab775a6ee6614f672a5f0f868677b523addf1d.

calypsomatic (approved):

app/new_import.py
nawatts: Added this comment to the code.

ashanhol (approved):
Is that code smell worth fixing? (not as familiar with python)

davidangb (approved):
thank you!

nawatts:
> Is that code smell worth fixing? (not as familiar with python)

> Return a value of type "ParseResult" instead of "tuple" or update function "validate_and_parse_url" type hint.
https://sonarcloud.io/project/issues?resolved=false&types=CODE_SMELL&pullRequest=138&id=broadinstitute_import-service&open=AYm8v-2XAF22vmHlLhcc

I think SonarCloud is wrong about this. `urlparse` does return a `ParseResult` (which is a named tuple, so maybe that's where SonarCloud is getting confused).
https://docs.python.org/3.9/library/urllib.parse.html#urllib.parse.urlparse
https://docs.python.org/3.9/library/urllib.parse.html#urlparse-result-object

PR #139, opened 2023-08-07
[AJ-1192] Configure allowed netlocs for imports per-environment

nawatts:
Import service only accepts import requests when the URL matches a specific set of domains (or more specifically, netlocs). Currently, this set of domains is hard-coded in new_import.py.

This moves that configuration to app.yaml, where the set of allowed domains can be configured based on the environment in which import service is deployed. Import service will also always accept imports from any GCS bucket, Azure storage container, or S3 bucket.

Also, checking valid netlocs is currently done by checking if the import URL ends with a valid netloc. Thus, if `example.com` is a valid netloc, then import service will accept imports from `example.com`, `subdomain.example.com`, `prefix-example.com`, etc. This changes that behavior to check if the request import URL's netloc exactly matches a configured valid netloc. The old behavior can be opted into by prefixing a valid netloc with `*`. This is used for allowing imports from Azure storage containers and S3 buckets, where the domain is specific to the storage container/bucket.

Finally, this updates the list of valid netlocs for Anvil. Currently, imports are allowed from all domains ending with either `anvil.gi.ucsc.edu` or `anvilproject.org`. This makes it so that, in production, imports are allowed from `service.prod.anvil.gi.ucsc.edu` and in other environments, imports are allowed from `service.anvil.gi.ucsc.edu`. Imports from `anvilproject.org` are no longer accepted.

### DIFF
--- a/app.yaml.ctmpl
+++ b/app.yaml.ctmpl
@@ -66,6 +66,8 @@ env_variables:
  RAWLS_PUBSUB_PROJECT: "broad-dsde-{{$env}}"
  RAWLS_PUBSUB_TOPIC: "rawls-async-import-topic-{{$env}}"
 
+ IMPORT_ALLOWED_NETLOCS: {{if eq $env "prod"}}"service.prod.anvil.gi.ucsc.edu,*dev.singlecell.gi.ucsc.edu,*gen3.biodatacatalyst.nhlbi.nih.gov,gen3-biodatacatalyst-nhlbi-nih-gov-pfb-export.s3.amazonaws.com,gen3-theanvil-io-pfb-export.s3.amazonaws.com,*service.azul.data.humancellatlas.org"
+                         {{else}}"service.anvil.gi.ucsc.edu,*dev.singlecell.gi.ucsc.edu,*gen3.biodatacatalyst.nhlbi.nih.gov,gen3-biodatacatalyst-nhlbi-nih-gov-pfb-export.s3.amazonaws.com,gen3-theanvil-io-pfb-export.s3.amazonaws.com,*service.azul.data.humancellatlas.org"{{end}}
 
 {{end}}{{end}}{{end}}{{end}}{{end}}{{end}}{{end}}{{end}}{{end}}
 
--- a/app.yaml.ctmpl
+++ b/app.yaml.ctmpl
@@ -66,8 +66,7 @@ env_variables:
  RAWLS_PUBSUB_PROJECT: "broad-dsde-{{$env}}"
  RAWLS_PUBSUB_TOPIC: "rawls-async-import-topic-{{$env}}"
 
- IMPORT_ALLOWED_NETLOCS: {{if eq $env "prod"}}"service.prod.anvil.gi.ucsc.edu,*dev.singlecell.gi.ucsc.edu,*gen3.biodatacatalyst.nhlbi.nih.gov,gen3-biodatacatalyst-nhlbi-nih-gov-pfb-export.s3.amazonaws.com,gen3-theanvil-io-pfb-export.s3.amazonaws.com,*service.azul.data.humancellatlas.org"
-                         {{else}}"service.anvil.gi.ucsc.edu,*dev.singlecell.gi.ucsc.edu,*gen3.biodatacatalyst.nhlbi.nih.gov,gen3-biodatacatalyst-nhlbi-nih-gov-pfb-export.s3.amazonaws.com,gen3-theanvil-io-pfb-export.s3.amazonaws.com,*service.azul.data.humancellatlas.org"{{end}}
+ IMPORT_ALLOWED_NETLOCS: "{{if eq $env "prod"}}service.prod.anvil.gi.ucsc.edu{{else}}service.anvil.gi.ucsc.edu{{end}},*dev.singlecell.gi.ucsc.edu,*gen3.biodatacatalyst.nhlbi.nih.gov,*service.azul.data.humancellatlas.org"
 
 {{end}}{{end}}{{end}}{{end}}{{end}}{{end}}{{end}}{{end}}{{end}}
 
--- a/app/new_import.py
+++ b/app/new_import.py
@@ -16,8 +16,22 @@ from app.auth.userinfo import UserInfo
 
 PROTECTED_NETLOCS = ["anvil.gi.ucsc.edu", "anvilproject.org", "gen3.biodatacatalyst.nhlbi.nih.gov", "gen3-biodatacatalyst-nhlbi-nih-gov-pfb-export.s3.amazonaws.com", "gen3-theanvil-io-pfb-export.s3.amazonaws.com"]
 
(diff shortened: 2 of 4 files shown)


app.yaml.ctmpl
```diff
@@ -66,6 +66,8 @@ env_variables:
  RAWLS_PUBSUB_PROJECT: "broad-dsde-{{$env}}"
  RAWLS_PUBSUB_TOPIC: "rawls-async-import-topic-{{$env}}"
 
+ IMPORT_ALLOWED_NETLOCS: {{if eq $env "prod"}}"service.prod.anvil.gi.ucsc.edu,*dev.singlecell.gi.ucsc.edu,*gen3.biodatacatalyst.nhlbi.nih.gov,gen3-biodatacatalyst-nhlbi-nih-gov-pfb-export.s3.amazonaws.com,gen3-theanvil-io-pfb-export.s3.amazonaws.com,*service.azul.data.humancellatlas.org"
+                         {{else}}"service.anvil.gi.ucsc.edu,*dev.singlecell.gi.ucsc.edu,*gen3.biodatacatalyst.nhlbi.nih.gov,gen3-biodatacatalyst-nhlbi-nih-gov-pfb-export.s3.amazonaws.com,gen3-theanvil-io-pfb-export.s3.amazonaws.com,*service.azul.data.humancellatlas.org"{{end}}
```
nawatts: We could remove the two S3 buckets from this list, since imports are always allowed from any S3 bucket.

app/new_import.py
```diff
@@ -123,12 +137,10 @@ def validate_import_url(import_url: Optional[str], import_filetype: Optional[str
 
     if import_filetype == FILETYPE_NOTRANSLATION and actual_netloc == os.environ.get("BATCH_UPSERT_BUCKET"):
         return actual_netloc
-    elif import_filetype in FILETYPE_TRANSLATORS.keys() and any(actual_netloc.endswith(s) for s in VALID_NETLOCS):
+    elif import_filetype in FILETYPE_TRANSLATORS.keys() and is_valid_netloc(parsedurl):
         return actual_netloc
     elif import_filetype == "tdrexport" and parsedurl.scheme == "gs":
         return actual_netloc
-    elif import_filetype == "tdrexport" and parsedurl.scheme == "https" and any(actual_netloc.endswith(s) for s in VALID_NETLOCS):
```
nawatts, line 130: This is a subset of the condition on line 140, so this branch was never hit.

app/tests/test_path_validation.py
```diff
@@ -56,13 +72,6 @@ def test_invalid_url(client, path, filetype):
     resp = client.post('/namespace/name/imports', json=payload, headers=good_headers)
     assert resp.status_code == 400
 
-@pytest.mark.parametrize("netloc", new_import.VALID_NETLOCS)
-@pytest.mark.usefixtures("sam_valid_user", "user_has_ws_access", "pubsub_publish", "pubsub_fake_env")
-def test_subdomain_of_legal_netlocs(client, netloc, caplog):
```
nawatts, line 61: Subdomains are nested in `test_default_valid_netlocs` and `test_configurd_valid_netlocs`.

app/new_import.py
```diff
@@ -123,12 +137,10 @@ def validate_import_url(import_url: Optional[str], import_filetype: Optional[str
 
     if import_filetype == FILETYPE_NOTRANSLATION and actual_netloc == os.environ.get("BATCH_UPSERT_BUCKET"):
         return actual_netloc
-    elif import_filetype in FILETYPE_TRANSLATORS.keys() and any(actual_netloc.endswith(s) for s in VALID_NETLOCS):
+    elif import_filetype in FILETYPE_TRANSLATORS.keys() and is_valid_netloc(parsedurl):
```
nawatts, line 140: I suspect that this should probably check the URL scheme as well.

app.yaml.ctmpl
davidangb: There's a bunch of duplication here; could you do something like this?
```
IMPORT_ALLOWED_NETLOCS: "{{if eq $env "prod"}}service.prod.anvil.gi.ucsc.edu{{else}}service.anvil.gi.ucsc.edu{{end}},*dev.singlecell.gi.ucsc.edu,*gen3.biodatacatalyst.nhlbi.nih.gov,gen3-biodatacatalyst-nhlbi-nih-gov-pfb-export.s3.amazonaws.com,gen3-theanvil-io-pfb-export.s3.amazonaws.com,*service.azul.data.humancellatlas.org"
```
I haven't tried this to see if there are any issues with whitespace or anything else.

davidangb: Also - this is lousy, but we'll need corresponding changes to support BEEs: https://github.com/broadinstitute/terra-helmfile/blob/master/values/app/importservice/bee.yaml.gotmpl

app/new_import.py
```diff
@@ -16,8 +16,22 @@
 
 PROTECTED_NETLOCS = ["anvil.gi.ucsc.edu", "anvilproject.org", "gen3.biodatacatalyst.nhlbi.nih.gov", "gen3-biodatacatalyst-nhlbi-nih-gov-pfb-export.s3.amazonaws.com", "gen3-theanvil-io-pfb-export.s3.amazonaws.com"]
```
davidangb: both `service.prod.anvil.gi.ucsc.edu` (prod) and `service.anvil.gi.ucsc.edu` (non-prod) need to be considered protected. Also we can remove `anvilproject.org` from the protected list since we won't allow imports from it (yet)

davidangb (commented):
see inline!

app/new_import.py
nawatts: These should still be protected from the inclusion of `anvil.gi.ucsc.edu` here. But yeah, it would be clearer if the netlocs in both places matched.

https://github.com/broadinstitute/import-service/blob/4f1ab5ebb941dd04f648c63eae47146f1058d34b/app/new_import.py#L136-L141

app.yaml.ctmpl
nawatts: Done in https://github.com/broadinstitute/import-service/pull/139/commits/4441f9655fc1932b084a8266b28833287b1420aa. Also removed the S3 buckets to tidy this up.

app/new_import.py
nawatts: Done in https://github.com/broadinstitute/import-service/pull/139/commits/566a5becd083e7c80b2674330706dfc45af133d3.

davidangb: ah, you're totally right - I didn't read closely enough! Thanks for the clarifying fix.

davidangb (approved):
did not test it myself, code changes look good - thank you!

app.yaml.ctmpl
nawatts: Opened https://github.com/broadinstitute/terra-helmfile/pull/4414 for the BEE configuration.

calypsomatic (approved):

PR #141, opened 2023-08-31
[AJ-1305] Prevent importing protected TDR snapshots into unprotected workspaces

nawatts:
Currently, when importing a TDR snapshot, import service ignores the snapshot's protected status ("secure monitoring enabled"), allowing a protected snapshot to be imported into an unprotected workspace. The desired behavior is for import service to require a workspace be protected (have an authorization domain or have all data access logged) in order to import a protected TDR snapshot into the workspace.

This change has import service load and parse the snapshot manifest file generated by TDR to determine if the import qualifies as protected data.

### DIFF
--- a/app/external/tdr_model.py
+++ b/app/external/tdr_model.py
@@ -18,6 +18,7 @@ class Dataset(BaseModel):
     defaultProfileId: str
     createdDate: str
     storage: List[StorageItem]
+    secureMonitoringEnabled: Optional[bool] = False
 
 
 class SourceItem(BaseModel):
--- a/app/external/tdr_model.py
+++ b/app/external/tdr_model.py
@@ -18,6 +18,7 @@ class Dataset(BaseModel):
     defaultProfileId: str
     createdDate: str
     storage: List[StorageItem]
+    # Older TDR manifests may not have this field, so made it optional when constructing a TDRManifest object.
     secureMonitoringEnabled: Optional[bool] = False
 
 
--- a/app/new_import.py
+++ b/app/new_import.py
@@ -1,11 +1,13 @@
 import flask
+import json
 import logging
 
 from app.translate import FILETYPE_TRANSLATORS, FILETYPE_NOTRANSLATION
 from app.db import db, model
-from app.external import sam, pubsub
+from app.external import gcs, sam, pubsub
+from app.external.tdr_model import TDRManifest
 from app.auth import user_auth
-from app.util import exceptions
+from app.util import exceptions, http
 
 from pydantic import AnyUrl, validate_arguments
 from typing import Optional, Set
@@ -64,7 +66,7 @@ def handle(request: flask.Request, ws_ns: str, ws_name: str) -> model.ImportStat
     # and validate the input's path
     validate_import_url(import_url, import_filetype, user_info)
     # Refuse to import protected data into unprotected workspace
-    if is_protected_data(import_url, import_filetype):
+    if is_protected_data(import_url, import_filetype, google_project=google_project, user_info=user_info):
         if not is_protected_workspace(authorization_domain, bucket_name):
             raise exceptions.AuthorizationException("Unable to import protected data into an unprotected workspace")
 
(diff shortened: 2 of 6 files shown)


app/external/tdr_model.py
```diff
@@ -18,6 +18,7 @@ class Dataset(BaseModel):
     defaultProfileId: str
     createdDate: str
     storage: List[StorageItem]
+    secureMonitoringEnabled: Optional[bool] = False
```
nawatts, line 22: Older TDR manifests may not have this field, so made it optional when constructing a TDRManifest object.

app/new_import.py
```diff
@@ -62,9 +64,9 @@ def handle(request: flask.Request, ws_ns: str, ws_name: str) -> model.ImportStat
         options: {options}, tdrSyncFlag: {is_tdr_sync_required}")
 
     # and validate the input's path
-    actual_netloc = validate_import_url(import_url, import_filetype, user_info)
+    validate_import_url(import_url, import_filetype, user_info)
     # Refuse to import protected data into unprotected workspace
-    if is_protected_data(actual_netloc, import_filetype):
+    if is_protected_data(import_url, import_filetype, google_project=google_project, user_info=user_info):
```
nawatts, line 69: `is_protected_data` now needs the full URL to the TDR manifest to download it.

It also requires some authentication information (`google_project` and `user_info`) in case the manifest is stored in GCS.

app/tests/conftest.py
```diff
@@ -98,7 +98,7 @@ def fake_pfb() -> Iterator[IO]:
 
 @pytest.fixture(scope="function")
 def fake_tdr_manifest() -> Iterator[IO]:
-    with open("app/tests/response_1638551384572.json", 'rb') as out:
```
nawatts, line 101: This file doesn't exist.

app/tests/resources/tdr_manifest_protected_dataset.json
nawatts, line 1: This file is mostly a copy of `test_tdr_response_gcp.json`, with `secureMonitoringEnabled: true` added to the source dataset.

app/new_import.py
```diff
@@ -145,9 +147,28 @@ def validate_import_url(import_url: Optional[str], import_filetype: Optional[str
         logging.warning(f"Unrecognized netloc or bucket for import: [{parsedurl.netloc}] from [{import_url}]")
         raise exceptions.InvalidPathException(import_url, user_info, "File cannot be imported from this URL.")
 
-def is_protected_data(import_netloc: str, import_filetype: Optional[str]) -> bool:
+def is_protected_data(import_url: str, import_filetype: str, *, google_project: str, user_info: UserInfo) -> bool:
```
calypsomatic, line 150: I'm unfamiliar with this `*` syntax!  A quick google suggests it means the following parameters require keyword syntax - does that mean they're not required?  Or why have it this way?

nawatts: Yes, it means that arguments for parameters after the `*` must be specified as keyword arguments.
https://peps.python.org/pep-3102/

As the number of positional parameters grows, it becomes harder to keep track of what's what. I think that using keywords, especially for arguments like these that are tangential to the main focus of the function, makes the calls easier to read and understand. But I have no issue removing this either.

calypsomatic: It's fine!  Just curious!

calypsomatic (approved):

app/external/tdr_model.py
jladieu: Might be worth including this very same comment in the code itself.  I think it was a helpful explanation!

app/new_import.py
jladieu: TIL, thanks for asking and sharing!

app/tests/resources/tdr_manifest_protected_dataset.json
jladieu: Does using `git cp` help at all to make the diff pathway clearer?

nawatts: I went to try it, but got:
```
git: 'cp' is not a git command.
```

app/tests/test_new_import.py
```diff
@@ -185,21 +191,24 @@ def test_validate_import_url(import_url, netloc, file_type_translator):
...
-@pytest.mark.parametrize("file_type", ["pfb", "tdrexport"])
-def test_is_protected_data(import_url, protected, file_type):
-    if file_type == "pfb":
-        assert is_protected_data(import_netloc=import_url, import_filetype=file_type) is protected
-    else:
-        assert is_protected_data(import_netloc=import_url, import_filetype=file_type) is False
+def test_is_protected_data_pfb(import_url, protected):
+    assert is_protected_data(import_url, "pfb", google_project="test_project", user_info=UserInfo("subject-id", "user@example.com", True)) is protected
 
+@pytest.mark.parametrize("manifest,protected", [
```
jladieu, line 205: I'm still getting familiar with parameterized tests, but given there are only two pathways here, it might actually be a little easier to read the tests if they were split out and standalone (and a little bit more duplicative), rather than having the conditional logic of whether or not `protected` is true.

app/external/tdr_model.py
nawatts: Added in https://github.com/broadinstitute/import-service/pull/141/commits/d9ef448f2483ae4020e3534f1db74b7976df9294.

app/tests/resources/tdr_manifest_protected_dataset.json
jladieu: Haha, oh weird....I've grown used to mercurial, which provided some of these extra options to preserve diff semantics.  Sorry for the goosechase!

jladieu (approved):

app/tests/test_new_import.py
nawatts: I think I'd prefer to keep them parameterized since broken up, it would be the same code but with different values in a few places, which seems like a good fit for `parametrize`.

I did split up the test for `is_protected_data` into separate tests for PFB and TDR exports, since the mocks in the TDR test are unnecessary for testing PFB exports.

PR #142, opened 2023-09-13
[AJ-1324] Allow imports from S3 using path style URLs

nawatts:
https://broadworkbench.atlassian.net/browse/AJ-1323

Import service is intended to accept imports from any S3 bucket. However, S3 supports multiple formats of URLs (https://docs.aws.amazon.com/AmazonS3/latest/userguide/VirtualHosting.html).

Currently, import service only supports the “virtual host style” “legacy global endpoint” (`https://<bucket>.s3.amazonaws.com/path/to/file`).

The HCA import is attempting to use the “path style” “legacy global endpoint” (`https://s3.amazonaws.com/<bucket>/path/to/file`). This adds support for it.

### DIFF
--- a/app/new_import.py
+++ b/app/new_import.py
@@ -15,8 +15,7 @@ from urllib.parse import ParseResult, urlparse
 import os
 
 from app.auth.userinfo import UserInfo
-
-PROTECTED_NETLOCS = ["service.prod.anvil.gi.ucsc.edu", "service.anvil.gi.ucsc.edu", "gen3.biodatacatalyst.nhlbi.nih.gov", "gen3-biodatacatalyst-nhlbi-nih-gov-pfb-export.s3.amazonaws.com", "gen3-theanvil-io-pfb-export.s3.amazonaws.com"]
+from app.protected_data import PROTECTED_URL_PATTERNS
 
 # Allow downloads from any GCS bucket, Azure storage container, or S3 bucket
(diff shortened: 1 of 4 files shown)


nawatts:
Some protected data sources are defined using the virtual host style S3 bucket URLs. The path style URLs for those buckets need to be added to the protected data sources list as well.

calypsomatic (approved):

davidangb (approved):

nawatts:
https://github.com/broadinstitute/import-service/pull/142/commits/af3a0cb6ef613ac2027863b68781a46f91db2019 updates the code to identify "protected" data to use the full URL instead of only the netloc/hostname. This allows us to identify path based S3 URLs as protected data.

https://github.com/broadinstitute/import-service/pull/142/commits/f4ed4805cc73a523aedac2ad47c898f04ee42a18 adds the AnVIL bucket mentioned in UCSC's message to the protected data list.

That AnVIL bucket still evades our check for importing production AnVIL data into dev Terra, but I am considering that as lower priority seeing as: 1) we didn't introduce the issue and 2) a user has to purposefully change a provided import URL to import production data into a dev environment (and if the user is purposefully trying to avoid that guard, then it's trivial to avoid anyway).

jladieu (approved):

calypsomatic (approved):

PR #143, opened 2023-09-22
[AJ-1324] Allow configuring restricted buckets

nawatts:
https://broadworkbench.atlassian.net/browse/AJ-1324

Stacked on #142.

This is part 2 of fixing the UCSC imports. By default, import service accepts imports from any GCS or S3 bucket or Azure storage container. After UCSC's changes, we need to support importing AnVIL data from S3 URLs. AnVIL data has the additional requirement that we block imports of production AnVIL data URLs into non-production Terra environments.

This adds an environment variable that allows configuring "restricted" S3 buckets, where imports from those buckets should be blocked. It configures the UCSC AnVIL bucket as a restricted source for non-production environments.

### DIFF
--- a/app/new_import.py
+++ b/app/new_import.py
@@ -74,7 +74,7 @@ def handle(request: flask.Request, ws_ns: str, ws_name: str) -> model.ImportStat
 
     # Refuse imports from restricted sources
     if protected_data.is_restricted_import(import_url):
-        raise exceptions.AuthorizationException("Unable to import data from this source")
+        raise exceptions.AuthorizationException("Unable to import data from this source into this Terra environment")
 
     # Refuse to import protected data into unprotected workspace
     if is_protected_data(import_url, import_filetype, google_project=google_project, user_info=user_info):
--- a/app/new_import.py
+++ b/app/new_import.py
@@ -15,7 +15,7 @@ from urllib.parse import ParseResult, urlparse
 import os
 
 from app.auth.userinfo import UserInfo
-from app.protected_data import PROTECTED_URL_PATTERNS
+from app import protected_data
 
 # Allow downloads from any GCS bucket, Azure storage container, or S3 bucket
 VALID_NETLOCS = [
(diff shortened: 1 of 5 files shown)


app/new_import.py
```diff
@@ -15,7 +15,7 @@
 import os
 
 from app.auth.userinfo import UserInfo
-from app.protected_data import PROTECTED_URL_PATTERNS
+from app import protected_data
```
jladieu, line 18: Nice!  I like this pattern and think you should also use it in `test_new_import.py`

app/tests/test_new_import.py
```diff
@@ -13,6 +13,7 @@
 from app.new_import import validate_import_url, is_protected_data, is_protected_workspace
 from app.util.exceptions import InvalidPathException
 from app.auth.userinfo import UserInfo
+from app.protected_data import url_patterns_for_s3_bucket
```
jladieu: I liked earlier how in `new_import.py` you switched the import to `from app import protected_data`, and then used `protected_data.<thing>` later on; I think that helps clarify inline where the keyword comes from and would be helpful here for tracing the definition of `url_patterns_for_s3_bucket` back.

app/new_import.py
```diff
@@ -71,6 +71,11 @@ def handle(request: flask.Request, ws_ns: str, ws_name: str) -> model.ImportStat
 
     # and validate the input's path
     validate_import_url(import_url, import_filetype, user_info)
+
+    # Refuse imports from restricted sources
+    if protected_data.is_restricted_import(import_url):
+        raise exceptions.AuthorizationException("Unable to import data from this source")
```
calypsomatic: This message seems a little generic - should we include any information about why?

jladieu (approved):

calypsomatic (approved):

app/tests/test_new_import.py
nawatts: 👍 Though I did that to allow mocking `RESTRICTED_URL_PATTERNS` in the test, not for clarity.

app/new_import.py
nawatts: I was thinking about this as a generic feature. But practically, it's probably not likely to expand in usage.

"Unable to import data from this source into this Terra environment"?

app/tests/test_new_import.py
nawatts: Done in https://github.com/broadinstitute/import-service/pull/143/commits/96f3235d985a677662b7dd9fe675fcb1dd0b86c2.

app/new_import.py
calypsomatic: Yeah that seems a bit better, generic but gives more of a clue

nawatts: Updated in https://github.com/broadinstitute/import-service/pull/143/commits/c1747582fd403df083b332cae9c25274fd377b62.

nawatts:
Rebased after merging #142.

PR #145, opened 2023-09-26
[AJ-1311] Block cross-platform TDR imports

nawatts:
https://broadworkbench.atlassian.net/browse/AJ-1311

Until the story for egress costs is figured out, we want to block imports of TDR snapshots across cloud platforms (GCP snapshot => Azure workspace, Azure snapshot => GCP workspace). This adds a check for cross-cloud platform imports to the initial import request validation.

### DIFF
--- a/app/new_import.py
+++ b/app/new_import.py
@@ -5,6 +5,7 @@ import logging
 from app.translate import FILETYPE_TRANSLATORS, FILETYPE_NOTRANSLATION
 from app.db import db, model
 from app.external import gcs, sam, pubsub
+from app.external.cloud_platform import CloudPlatform
 from app.external.tdr_model import TDRManifest
 from app.auth import user_auth
 from app.util import exceptions, http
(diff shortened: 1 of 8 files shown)


app/new_import.py
```diff
@@ -187,3 +191,10 @@ def load_tdr_manifest(manifest_url: str, *, google_project: str, user_info: User
 def is_protected_snapshot(manifest: TDRManifest) -> bool:
     """Determine whether a TDR manifest contains protected data."""
     return any(source.dataset.secureMonitoringEnabled for source in manifest.snapshot.source)
+
+def all_sources_on_cloud_platform(manifest: TDRManifest, cloud_platform: CloudPlatform) -> bool:
```
nawatts, line 195: I wasn't entirely sure how to interpret multiple storage items per dataset and multiple datasets per snapshot in this context. I fell back on requiring that all storage items listed in the manifest must be on the same cloud platform as the workspace.

@nmalfroy does this make sense?

app/tests/test_new_import.py
```diff
@@ -233,3 +233,17 @@ def test_restricted_imports(client):
...
+    ("test_tdr_response_azure.json", "gcp"),
+])
+@pytest.mark.usefixtures("sam_valid_user", "pubsub_publish", "pubsub_fake_env")
+def test_blocks_cross_platform_tdr_imports(monkeypatch, client, manifest, workspace_cloud_platform):
+    monkeypatch.setattr("app.auth.user_auth.workspace_uuid_and_project_with_auth",
+                        mock.MagicMock(return_value=RawlsWorkspaceResponse("aaaaaaaa-bbbb-cccc-dddd-eeeeeeeeeeee", "some-project", workspace_cloud_platform, set(), "fc-11111111-2222-3333-4444-555555555555")))
+    monkeypatch.setattr("app.new_import.http.http_as_filelike", mock.MagicMock(return_value=open(f"app/tests/resources/{manifest}", "rb")))
+
+    resp = client.post('/mynamespace/myname/imports', json=good_tdr_json, headers=good_headers)
+    assert resp.status_code == 403
```
nawatts, line 248: This is the same error code we've been using to reject importing protected data into unprotected workspaces. But I think we're misusing it and this and other import validation should return 400.

davidangb: I agree, 400 would be better. I wouldn't block the PR on this if that turns into a big deal.

app/new_import.py
davidangb: this sounds right to me!

davidangb (approved):
thank you!

app/tests/test_new_import.py
nawatts: Since import service is already sending 403s for similar cases, I'll change this in a follow-up PR.

jladieu (approved):

app/tests/test_new_import.py
nawatts: Changed in #147.

PR #146, opened 2023-09-26
DR-3210 Allow imports from google signed urls from TDR

nmalfroy:
We've seen issues with permission propagation when importing from `gs://` paths.  That's super annoying! This PR allows manifests and parquet files in TDR exports to be signed URLs.  The only thing here is that the urls might expire but that seems more tractable than not knowing when the permissions will become available.  Also, this copies a pattern that we are already using for Azure files (if we ever decide to re-allow importing an azure snapshot into a Google workspace)

A couple of drive-bys:
- Because the likelihood of expiring links, we no longer throw a `FileTooBigToDownload` error when accessing a file via https throws a 400.  We now throw an `InvalidFileUrl` (a little vague but not outright incorrect)
- Refactored some unit tests to be able to have more parameterized tests

### DIFF
--- a/app/translators/tdr_manifest_to_rawls.py
+++ b/app/translators/tdr_manifest_to_rawls.py
@@ -98,14 +98,14 @@ class ParquetTranslator:
             bucket = parsedurl.netloc
             path = parsedurl.path
             with gcs.open_file(self.import_details.workspace_google_project, bucket, path, self.import_details.submitter, self.auth_key) as pqfile:
-                return self.translate_parquet_file_to_entities(pqfile, is_azure=False, ref_only=ref_only)
+                return self.translate_parquet_file_to_entities(pqfile, is_https=False, is_azure=False, ref_only=ref_only)
         elif (parsedurl.scheme == 'https'):
             hostname = parsedurl.netloc
             if not (hostname.endswith(VALID_AZURE_DOMAIN) or hostname == GOOGLE_STORAGE_DOMAIN):
                 logging.error(f"unsupported domain in url {self.filelocation} provided")
                 raise exceptions.InvalidPathException(self.filelocation, user_info, "Unsupported domain")
             with http.http_as_filelike(self.filelocation) as pqfile:
-                return self.translate_parquet_file_to_entities(pqfile, is_azure=hostname.endswith(VALID_AZURE_DOMAIN), ref_only=ref_only)
+                return self.translate_parquet_file_to_entities(pqfile, is_https=True, is_azure=hostname.endswith(VALID_AZURE_DOMAIN), ref_only=ref_only)
         else:
             logging.error(f"unsupported scheme {parsedurl.scheme} provided")
             raise exceptions.InvalidPathException(self.filelocation, user_info, "Unsupported scheme")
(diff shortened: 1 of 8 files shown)


davidangb (approved):
the code as-is lgtm, so I am giving a :+1: … but I wonder if this will have any interactions with #145.

app/util/exceptions.py
```diff
@@ -53,6 +53,10 @@ def __init__(self, message: str = "File too large to download or did not report
         """Thrown when we detect a file is dangerously large."""
         super().__init__(message, 413)
 
+class InvalidFileUrl(ISvcException):
+    def __init__(self, message: str = "The file URL was invalid. It is possible that the URL has expired"):
```
nawatts: Extreme nit: since the first sentence in the message uses punctuation, the second should as well.

```suggestion
    def __init__(self, message: str = "The file URL was invalid. It is possible that the URL has expired."):
```

app/translators/tdr_manifest_to_rawls.py
```diff
@@ -100,12 +101,11 @@ def translate(self, ref_only: bool = False) -> Iterator[Entity]:
...
         elif (parsedurl.scheme == 'https'):
             hostname = parsedurl.netloc
-            if not hostname.endswith(VALID_AZURE_DOMAIN):
+            if not (hostname.endswith(VALID_AZURE_DOMAIN) or hostname == GOOGLE_STORAGE_DOMAIN):
                 logging.error(f"unsupported domain in url {self.filelocation} provided")
                 raise exceptions.InvalidPathException(self.filelocation, user_info, "Unsupported domain")
             with http.http_as_filelike(self.filelocation) as pqfile:
-                # Assumption is that if a file is using https, it's an Azure snapshot
-                return self.translate_parquet_file_to_entities(pqfile, True, ref_only)
+                return self.translate_parquet_file_to_entities(pqfile, hostname.endswith(VALID_AZURE_DOMAIN), ref_only)
```
nawatts: Nit: Using a keyword argument for `is_azure` would make this a little clearer.

```suggestion
                return self.translate_parquet_file_to_entities(pqfile, is_azure=hostname.endswith(VALID_AZURE_DOMAIN), ref_only=ref_only)
```

nawatts (approved):

nawatts:
> but I wonder if this will have any interactions with https://github.com/broadinstitute/import-service/pull/145.

Possibly some conflicts in tests to resolve but, otherwise, #145 is confined to the initial import request / validation and this is dealing with the following translate request.

app/util/exceptions.py
nmalfroy: Haha I actually went back and forth on this

app/translators/tdr_manifest_to_rawls.py
nmalfroy: ooh good call.  I'm not super used to python so this is super good to know

nmalfroy:
> > but I wonder if this will have any interactions with #145.
> 
> Possibly some conflicts in tests to resolve but, otherwise, #145 is confined to the initial import request / validation and this is dealing with the following translate request.

Are you ok if I end up merging?  I'm happy to rebase if you want to go first

nawatts:
Merge away! I still need one more review on #145.

app/translators/tdr_manifest_to_rawls.py
```diff
@@ -118,10 +118,10 @@ def translate(self, ref_only: bool = False) -> Iterator[Entity]:
     #     assert reader.schema == 'foo'
     #     entity_batches = (self.translate_data_frame(b.to_pandas(), b.column_names) for b in reader)
     #     return itertools.chain(*entity_batches)
-    def translate_parquet_file_to_entities(self, file_like: IO, is_azure: bool = False, ref_only: bool = False) -> Iterator[Entity]:
+    def translate_parquet_file_to_entities(self, file_like: IO, is_azure: bool = False, is_https: bool = False, ref_only: bool = False) -> Iterator[Entity]:
         """Converts single parquet file-like object to an iterator of Entity objects."""
         pq_table: pyarrow.Table
-        if (is_azure):
+        if (is_azure or is_https):
```
nawatts, line 124: Could this be simplified to `is_https`? Is `is_azure` ever true when `is_https` is not?

nmalfroy: I was debating.  I ran into issues trying to conflate the two so it seemed reasonable to keep both checks to be safe

PR #154, opened 2023-10-18
[AJ-1399] Add --no-cache flag to gcloud app deploy

nawatts:
Since we put a lifecycle rule on the artifacts bucket that deletes objects after 7 days, older images in the Container Registry may be broken since their underlying storage objects may have been deleted.

This attempts to work around that by telling the App Engine deploy step not to use a cache.

### DIFF
--- a/deploy.sh
+++ b/deploy.sh
@@ -74,4 +74,4 @@ docker run -v $PWD:/app \
   -w /app \
   --entrypoint "/bin/bash" \
    $CLOUD_SDK_DOCKER_IMG \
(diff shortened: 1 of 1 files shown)


calypsomatic (approved):

davidangb (approved):
Remember that we're on an old version of `gcloud` (https://broadinstitute.slack.com/archives/CADM7MZ35/p1695908807120359?thread_ts=1695668221.683489&cid=CADM7MZ35) … if you run into a problem with this new flag we may need to update

nawatts:
The deploy successfully ran on this commit, so no issues with the flag.

We'll have to wait for a week without a deploy to see if it resolves the issue related to the lifecycle rule.

PR #159, opened 2023-11-16
Bump to workzeug 2.3.8 to fix cve-2023-46136

jladieu:
Release notes for workzeug [2.3.8](https://github.com/pallets/werkzeug/blob/main/CHANGES.rst#version-238)

Alternative to https://github.com/broadinstitute/import-service/pull/155 which bumps workzeug but uses the next major version which creates a bunch of dependency conflicts.


nawatts:
> Use a direct github reference as the package is not yet published where Poetry can see it.

? There's a version 2.3.8 on PyPI: https://pypi.org/project/Werkzeug/2.3.8/

davidangb (commented):
I see 2.3.8 on pypi: https://pypi.org/project/Werkzeug/2.3.8/ … is that not working?

You will sadly also have to update requirements.txt - see instructions at https://github.com/broadinstitute/import-service?tab=readme-ov-file#dependency-management

jladieu:
Alright, I'm not sure why I couldn't find 2.3.8 when I was working on this yesterday, but I managed to get it working today after going on a detour which resulted in the extra snippet in the README.md.

davidangb (approved):
perhaps add a comment in pyproject.toml that werkzeug is a transitive dependency we needed to udpate? Otherwise looks good.

sarahgibs (approved):

